PR #52, opened 2021-09-19
Change to yt-dlp, Added settings dialog with downloading presets

itsteddyyo:
The main change is to replace youtube-dl with yt-dlp. It is a fork of youtube-dl (https://github.com/yt-dlp/yt-dlp) that has the upstream merged regularly. Plus it has many neat features like the --throttled-rate option used in my changes. In my experience the download speed without using any command line options is faster too. 
Essentially it is just a little more alive than youtube-dl. The developer is also always fast to respond to issues.

I also added a settings dialog. It is accesible from the normal right-click menu.
image

There are options for setting the executable paths and some yt-dlp options.
image

1. Download Retry Speed: The mentioned --throttled-rate. It essentially restarts the download if the speed is long enough under the threshold. The already downloaded parts are kept and not redownloaded. So you essentially have no downside.
2. Quality options: Here are some presets for downloading youtube videos.
image

- yt-dlp Default: is the default with no command-line arguments (apart from default 'best'). Downloads only up 720p. That is because youtube-dl and yt-dlp initially only download videos with video and audio in one file. But for videos with 1080p and up Youtube interestinly stores the video and audio in two distinct files. So you need ffmpeg/ffprobe or any other tool that does the same job to put audio and video together :D

- Better: downloads video up to 1080p + the best available audio format and puts the two together.
- Best: downloads video with best resolution and quality + the best available audio format and puts the two together.

.mp4 is prefered over .webm by default so conversion should not be neeeded in most cases.

Those are just some that made the most sense too me and looked good while trying. If users have some cool options those could be added too :D The default preset at the beginning is 'Better' with the 1080p. The default throttle rate is 1mb/s.

You can see which options are used by the presets when you select one, save and then open the window again.
The command line options used are -f and -S for filtering and sorting videos. The documentation is linked (only text as i could not figure out how to make proper links :D)

There is also an option 'Manual' that enables the textboxes (disabled normally when a preset is selected) for input and let's the user filter/sort whatever he wants at his own risk. But then again you can always go back to the presets.

In the code there are quite some changes in the youtube section (steam download was not touched). So here just the most important:
1. Invoke-ViewSettings for already mentioned settings dialogue
2. Distinct Download-YoutubeVideo function that is used for both manually and automatically downloading.
3. Distinct Get-YTDLPSettings function that concatenates the command line argument and correctly reacts to empty settings
4. The list from the manual downloading now gets populated by the result of yt-dlp too so the settings the user wants are represented here too. Plus i added the resolution in the list window.

I hope everything is formatted correctly. It looks good to me but then again you become blind too your own code after a few days. I also tested all the presets and they worked for me so i hope everythings good there too.

I didn't squash the commits as i did not know if you would want the whole history for reviewing. If you want that changed just let me know.

Already big thanks upfront for taking the time to look at my changes :D

### DIFF
--- a/source/Generic/ReviewViewer/Controls/ReviewsControl.xaml
+++ b/source/Generic/ReviewViewer/Controls/ReviewsControl.xaml
@@ -92,15 +92,18 @@
             <ScrollViewer VerticalScrollBarVisibility="Auto" Margin="0,0,0,10" Height="{Binding SettingsModel.Settings.DescriptionHeight}">
                 <TextBlock Text="{Binding SelectedReview.ReviewReview}" TextWrapping="Wrap" Opacity="0.75"/>
             </ScrollViewer>
+            <StackPanel Margin="0,0,0,10" Visibility="{Binding SettingsModel.Settings.DisplayHelpfulnessData, Converter={StaticResource BooleanToVisibilityConverter}}">
+                <TextBlock Text="{Binding ReviewHelpfulnessHelpful}" Opacity="0.75" />
+                <TextBlock Text="{Binding ReviewHelpfulnessFunny}" Opacity="0.75" />
+            </StackPanel>
         </StackPanel>
-        
         <DockPanel>
             <StackPanel>
                 <Button Content="{DynamicResource LOCReview_Viewer_ReviewTypeAll}" MinWidth="60" Margin="5" Command="{Binding SwitchAllReviewsCommand}"/>
                 <Button Content="{DynamicResource LOCReview_Viewer_ReviewTypePositive}" MinWidth="60" Margin="5" Command="{Binding SwitchPositiveReviewsCommand}"/>
                 <Button Content="{DynamicResource LOCReview_Viewer_ReviewTypeNegative}" MinWidth="60" Margin="5" Command="{Binding SwitchNegativeReviewsCommand}"/>
             </StackPanel>
-            <Grid Name="SummaryGrid" Margin="0,0,0,10" Visibility="{Binding MainPanelVisibility, Mode=OneWay}">
+            <Grid Name="SummaryGrid" Margin="0,0,0,10" Visibility="{Binding MainPanelVisibility, Mode=OneWay}" Opacity="0.75">
                 <Grid.ColumnDefinitions>
                     <ColumnDefinition Width="Auto" />
                     <ColumnDefinition Width="Auto" />
--- a/source/Generic/ReviewViewer/Controls/ReviewsControl.xaml.cs
+++ b/source/Generic/ReviewViewer/Controls/ReviewsControl.xaml.cs
@@ -136,6 +136,28 @@ namespace ReviewViewer.Controls
             }
         }
 
+        private string reviewHelpfulnessHelpful;
+        public string ReviewHelpfulnessHelpful
+        {
+            get => reviewHelpfulnessHelpful;
+            set
+            {
+                reviewHelpfulnessHelpful = value;
+                OnPropertyChanged();
+            }
+        }
+
+        private string reviewHelpfulnessFunny;
+        public string ReviewHelpfulnessFunny
+        {
+            get => reviewHelpfulnessFunny;
+            set
+            {
+                reviewHelpfulnessFunny = value;
+                OnPropertyChanged();
+            }
+        }
+
         private int selectedReviewIndex;
         public int SelectedReviewIndex
         {
@@ -145,23 +167,20 @@ namespace ReviewViewer.Controls
                 selectedReviewIndex = value;
                 OnPropertyChanged();
                 SelectedReview = reviews.Reviews[selectedReviewIndex];
-                SetFormattedPlaytime();
-                SetTotalFormattedPlaytime();
+                SetReviewInfo();
             }
         }
 
-        private void SetFormattedPlaytime()
+        private void SetReviewInfo()
         {
             var time = TimeSpan.FromMinutes(SelectedReview.Author.PlaytimeAtReview);
-            var minutes = time.ToString("mm");
-            FormattedPlaytime = string.Format(ResourceProvider.GetString("LOCReview_Viewer_ReviewPlaytimeFormat"), Math.Floor(time.TotalHours), minutes);
-        }
+            FormattedPlaytime = string.Format(ResourceProvider.GetString("LOCReview_Viewer_ReviewPlaytimeFormat"), Math.Floor(time.TotalHours), time.ToString("mm"));
 
-        private void SetTotalFormattedPlaytime()
-        {
-            var time = TimeSpan.FromMinutes(SelectedReview.Author.PlaytimeForever);
-            var minutes = time.ToString("mm");
-            TotalFormattedPlaytime = string.Format(ResourceProvider.GetString("LOCReview_Viewer_ReviewTotalPlaytimeFormat"), Math.Floor(time.TotalHours), minutes);
+            time = TimeSpan.FromMinutes(SelectedReview.Author.PlaytimeForever);
+            TotalFormattedPlaytime = string.Format(ResourceProvider.GetString("LOCReview_Viewer_ReviewTotalPlaytimeFormat"), Math.Floor(time.TotalHours), time.ToString("mm"));
+
+            ReviewHelpfulnessHelpful = string.Format(ResourceProvider.GetString("LOCReview_Viewer_ReviewHelpfulnessHelpfulFormat"), SelectedReview.VotesUp);
+            ReviewHelpfulnessFunny = string.Format(ResourceProvider.GetString("LOCReview_Viewer_ReviewHelpfulnessFunnyFormat"), SelectedReview.VotesFunny);
         }
 
         private Review selectedReview;
--- a/source/Generic/ReviewViewer/Localization/en_US.xaml
+++ b/source/Generic/ReviewViewer/Localization/en_US.xaml
@@ -8,6 +8,7 @@
     <sys:String x:Key="LOCReview_Viewer_SettingUseMatchingSteamApiLang">Use language used by Playnite to search for reviews if it's available on Steam</sys:String>
     <sys:String x:Key="LOCReview_Viewer_SettingDownloadOnGameSelectionDescription">Download review data on game selection if it's missing</sys:String>
     <sys:String x:Key="LOCReview_Viewer_SettingDescriptionHeightDescription">Description height</sys:String>
+    <sys:String x:Key="LOCReview_Viewer_SettingDisplayHelpfulnessDataDescription">Display review helpfulness</sys:String>
     <sys:String x:Key="LOCReview_Viewer_TotalUserReviews">Total reviews:</sys:String>
     <sys:String x:Key="LOCReview_Viewer_PositiveUserReviews">Positive reviews:</sys:String>
     <sys:String x:Key="LOCReview_Viewer_NegativeUserReviews">Negative reviews:</sys:String>
@@ -24,4 +25,6 @@
     <sys:String x:Key="LOCReview_Viewer_ReviewNotRecommendedLabel">Not recommended</sys:String>
     <sys:String x:Key="LOCReview_Viewer_ReviewPlaytimeFormat">{0} hours {1} minutes at review time</sys:String>
     <sys:String x:Key="LOCReview_Viewer_ReviewTotalPlaytimeFormat">{0} hours {1} minutes of total play time</sys:String>
+    <sys:String x:Key="LOCReview_Viewer_ReviewHelpfulnessHelpfulFormat">{0} people found this review helpful</sys:String>
+    <sys:String x:Key="LOCReview_Viewer_ReviewHelpfulnessFunnyFormat">{0} people found this review funny</sys:String>
 </ResourceDictionary>
--- a/source/Generic/ReviewViewer/ReviewViewerSettings.cs
+++ b/source/Generic/ReviewViewer/ReviewViewerSettings.cs
@@ -45,6 +45,18 @@ namespace ReviewViewer
                 OnPropertyChanged();
             }
         }
+
+        [DontSerialize]
+        private bool displayHelpfulnessData { get; set; } = true;
+        public bool DisplayHelpfulnessData
+        {
+            get => displayHelpfulnessData;
+            set
+            {
+                displayHelpfulnessData = value;
+                OnPropertyChanged();
+            }
+        }
     }
 
     public class ReviewViewerSettingsViewModel : ObservableObject, ISettings
--- a/source/Generic/ReviewViewer/ReviewViewerSettingsView.xaml
+++ b/source/Generic/ReviewViewer/ReviewViewerSettingsView.xaml
@@ -10,6 +10,8 @@
                   Content="{DynamicResource LOCReview_Viewer_SettingUseMatchingSteamApiLang}" Margin="0,0,0,10"/>
         <CheckBox IsChecked="{Binding Settings.DownloadDataOnGameSelection}"
                   Content="{DynamicResource LOCReview_Viewer_SettingDownloadOnGameSelectionDescription}" Margin="0,0,0,10"/>
+        <CheckBox IsChecked="{Binding Settings.DisplayHelpfulnessData}"
+                  Content="{DynamicResource LOCReview_Viewer_SettingDisplayHelpfulnessDataDescription}" Margin="0,0,0,10"/>
         <StackPanel Orientation="Horizontal" Margin="0,0,0,10">
             <Label Content="{DynamicResource LOCReview_Viewer_SettingDescriptionHeightDescription}" Margin="0,0,0,0" VerticalAlignment="Center" />
             <Slider Value="{Binding Settings.DescriptionHeight, UpdateSourceTrigger=PropertyChanged}" Width="150" Minimum="50" Maximum="400" VerticalAlignment="Center"


source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
```diff
@@ -967,9 +1183,10 @@ function Invoke-YoutubeSearchWindow
                         <Image Margin="3" Height="80" Source="{Binding Thumbnail}"/>
                         <StackPanel Margin="5,0,0,0">
                             <TextBlock Margin="3" Text="{Binding Name}"/>
+                            <TextBlock Margin="0" Text="{Binding ChannelName}"/>
```
darklinkpower, line 1186: Here it should have the same margin or at least left margin as the other textblocks for alignment.

I think all left margins should be removed and just be set from the stack panel to `8`, only keeping the margins for below, like `Margin="0,0,0,3"`

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
```diff
@@ -59,15 +59,241 @@ function GetGameMenuItems
     $menuItem11.FunctionName = "Update-AssetsStatusGameDatabase"
     $menuItem11.MenuSection = "Extra Metadata tools"
 
-    return $menuItem, $menuItem2, $menuItem3, $menuItem9, $menuItem10, $menuItem4, $menuItem5, $menuItem6, $menuItem7, $menuItem8, $menuItem11
+    $menuItem12 = New-Object Playnite.SDK.Plugins.ScriptGameMenuItem
+    $menuItem12.Description = [Playnite.SDK.ResourceProvider]::GetString("LOCExtra_Metadata_tools_MenuItemInvoke-ViewSettingsDescription")
+    $menuItem12.FunctionName = "Invoke-ViewSettings"
+    $menuItem12.MenuSection = "Extra Metadata tools"
```
darklinkpower, line 65: I think the menu item should be in the video section since it's specific to that section only. Other option is to add the word "Video" to the settings description, whichever you prefer.

darklinkpower:
I'm getting this window that I have to close manually when searching for videos manually

image

darklinkpower:
Selecting the 720p option doesn't get saved and sets the quality to `Manual`

image


https://user-images.githubusercontent.com/1389286/134068268-184ac4c4-7598-48c9-8de2-d54894d60e83.mp4

darklinkpower:
This string should be changed to reflect the change to `yt-dlp`

image

darklinkpower:
Search now has issues with special characters or letters with accents or other punctuation

Now:

image

Before:

Playnite DesktopApp_COnzi7iz3v

Playnite DesktopApp_JFO3rlBD62

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
```diff
@@ -1049,29 +1266,74 @@ function Set-YouTubeVideoManual
...
-    
+
+    $ytdlp = $settings.ytdlpPath
+    [string]$ytdlpSettings = Get-YTDLPSettings
+
+    $videoPathNoExtension = Join-Path $extraMetadataDirectory -ChildPath "VideoTrailer."
+    $videoTempPathNoExtension = Join-Path $extraMetadataDirectory -ChildPath "VideoTemp."
+
+
+    $filePresent = ((Get-ChildItem -Path "$videoPathNoExtension*").Count -ne 0)
```
darklinkpower, line 1299: I don't fully understand why it's using GCI here, can you explain please?

ExtraMetadataTools video loading only seeks `VideoTrailer.mp4` inside the game EMT directory.

For that reason, it should be using the `Test-Path` cmdlet or the `Exists` method from the net File class

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
```diff
@@ -814,98 +1070,58 @@ function Set-YouTubeVideo
...
         [string]$queryInput
     )
+    
+    $settings = Get-Settings
 
-    $query = [uri]::EscapeDataString($queryInput)
-    $uri = "https://www.youtube.com/results?search_query={0}&sp=EgQQARgB" -f $query
-    $webContent = Get-DownloadString $uri
-    $webContent -match 'var ytInitialData = ((.*?(?=(;<\/script>))))' | Out-Null
+    $search = "`"ytsearch20:{0}`"" -f $queryInput
```
darklinkpower, line 1091: Just to make sure, since you are passing the raw input to search now. 

Could there be potential issues if you don't escape the string or issues with certain characters?

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
```diff
@@ -1049,29 +1266,74 @@ function Set-YouTubeVideoManual
...
+        'FilePath'     = $ytdlp
+        'ArgumentList' = @("-o `"$videoTempPathNoExtension%(ext)s`"", $ytdlpSettings, $search)
         'Wait'         = $true
         'PassThru'     = $true
     }
-    $proc = Start-Process @trailerdownloadparams -WindowStyle $settings.youtubedlWindowStyle
+    
+    $proc = Start-Process @trailerdownloadparams -WindowStyle $settings.ytdlpWindowStyle
+
+    $fileTemp = Get-ChildItem -Path "$videoTempPathNoExtension*" | Select BaseName, Extension
```
darklinkpower, line 1324: Minor thing, but can you change `Select` alias to the full function name please? `Select-Object`

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
```diff
@@ -59,15 +59,241 @@ function GetGameMenuItems
...
+        [PSCustomObject]@{
+            Name = "Manual"
+            Value = "manual"
+            Filter = ""
+            Sort = ""
+        }
+    )
+
+    #Get selected preset by comparing filter + sort
+    $selectedList = $QualityModeSource | Where { $_.Filter -eq $settings.filterFormat } | Where { $_.Sort -eq $settings.sortFormat }
```
darklinkpower, line 192: Same, change the `Where` alias to the full function name `Where-Object` please

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
```diff
@@ -59,15 +59,241 @@ function GetGameMenuItems
...
+    <DockPanel Margin="20">
+        <DockPanel DockPanel.Dock="Bottom" LastChildFill="False">
+            <Button Name="ButtonCancel" Margin="10,0,0,0" DockPanel.Dock="Right" VerticalAlignment="Bottom"/>
+            <Button Name="ButtonSave" Margin="0,0,0,0" DockPanel.Dock="Right" VerticalAlignment="Bottom"/>
+        </DockPanel>
+        <StackPanel Grid.Row="0" DockPanel.Dock="Top">
+
+            <TextBlock Text="{DynamicResource LOCExtra_Metadata_tools_FFMPEGP_PATH}" Name="TBLffmpegPath" Margin="0,0,0,3"/>
+            <DockPanel Margin="0,0,0,10">
+                <Button Name="BTNffmpegPath" Margin="10,0,0,0" DockPanel.Dock="Right">...</Button>
```
darklinkpower, line 98: In all buttons, please just set the Content just by naming the property like `Content="..."` instead of opening
In this case, it would be
`<Button Name="BTNffmpegPath" Margin="10,0,0,0" DockPanel.Dock="Right" Content="..." />`

itsteddyyo: Done in 04a0d39d82237c61cbdc29853f1fe31fe68bbe30

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
itsteddyyo: Done in 30b385caf6ed9eef73af37eaed397e5b316ceced

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
itsteddyyo: Done in 30b385caf6ed9eef73af37eaed397e5b316ceced

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
itsteddyyo: You are right . That is a leftover from checking if .webm without converting would be ok.
Done in 51bbcf88bcea825a6ffb414ed668b2ac5ec39ad1

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
itsteddyyo: Moved to 'Video' menu.
Done in 8718cdfc5830cdd1bb187cd65b3c2bad02bbde44

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
itsteddyyo: I hope i understood you right there. Just let me know if that was wrong.
Done in cb8701e19ceef61080e5c66542ce44b05b026f96

itsteddyyo:
> Search now has issues with special characters or letters with accents or other punctuation
> 
> Now:
> 
> image
> 
> Before:
> 
> Playnite DesktopApp_COnzi7iz3v
> 
> Playnite DesktopApp_JFO3rlBD62

Done in cd64147ad47bb80f6bcd1ceff262d932ff8e24a3.
Fixed for most characters. It currently does not work for special characters like the heart one in your example. Would that be good enough or should i keep looking?

itsteddyyo:
> This string should be changed to reflect the change to `yt-dlp`
> 
> image

Done in a51a8f13017fe1e97a82677ce4b670ae72e48299

itsteddyyo:
> Selecting the 720p option doesn't get saved and sets the quality to `Manual`
> 
> image
> 
>  vQdsUaPb1u.mp4

Done in 6396b8f294a961c4dc26f11adbf51fe1db74d4dc.
Now loading of empty strings from the settings is allowd

itsteddyyo:
> I'm getting this window that I have to close manually when searching for videos manually
> 
> image
> 
> Probably related but sometimes there are only very few results
> 
> image

Removed the 0,1,2 output in abe2a1c1297887a3da481650bf308a02fe902b93.
But i could not recreate the issue with the window not closing. Plus i always have full results.
What yt-dlp ffmpeg and ffprobe versions are you using?
Could you maybe retry with the new changes?

source/Generic/ExtraMetadataToolsVideo/ExtraMetadataToolsVideo.psm1
itsteddyyo: Yeah you're right sorry. This seems not to be so straightforward. I found this (https://gist.github.com/bilbothebaggins/efbbc2312ec72c4967622b340e4b26d6) and it works. But i requires to have that code to be added to the project which isn't too great. An easier solution would be much better. Would you happen to have any ideas? 😅

itsteddyyo: Very rudimentary done in d62f860bcb72053ecc9797252c03ea01790ec09b by removing quotes from search string.
Not perfect but then again it is stable enough for normal use cases and if someone wants to forcefully wreck their query then goodspeed to them. After all were talking a dialogue to download videos in Playnite here and not some vulnerable banking software.

darklinkpower:
> Removed the 0,1,2 output in [abe2a1c](https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/abe2a1c1297887a3da481650bf308a02fe902b93).
> But i could not recreate the issue with the window not closing. Plus i always have full results.
> What yt-dlp ffmpeg and ffprobe versions are you using?
> Could you maybe retry with the new changes?

Correct me if I'm wrong but I think both ffprobe and ffmpeg are irrelevant for results searching and could not affect it?
As for yt-dlp I downloaded the latest release from https://github.com/yt-dlp/yt-dlp/releases/tag/2021.09.02

darklinkpower:
> ...Done in [cd64147](https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/cd64147ad47bb80f6bcd1ceff262d932ff8e24a3).
> Fixed for most characters. It currently does not work for special characters like the heart one in your example. Would that be good enough or should i keep looking?
> 
> Now completly fixed by switching to --dump-json in [f989da1](https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/f989da18c4c9256b8eb88a0bdd0cb228a3a941b0).

Regarding the search results, just wondering what are the benefits of searching with yt-dlp as opposed to how it was done before by manually parsing the results search page. I see that doing it this way is causing some issues, needs some extra steps like outputting to a local file, parsing, etc. and could be more error prone in certain circumstances, which makes me a little unsure if this is a needed or good change.

> The list from the manual downloading now gets populated by the result of yt-dlp too so the settings the user wants are represented here too. Plus i added the resolution in the list window.

I'm not sure what settings should be affecting the search, can you elaborate please?

itsteddyyo:
> > Removed the 0,1,2 output in [abe2a1c](https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/abe2a1c1297887a3da481650bf308a02fe902b93).
> > But i could not recreate the issue with the window not closing. Plus i always have full results.
> > What yt-dlp ffmpeg and ffprobe versions are you using?
> > Could you maybe retry with the new changes?
> 
> Correct me if I'm wrong but I think both ffprobe and ffmpeg are irrelevant for results searching and could not affect it?
> As for yt-dlp I downloaded the latest release from https://github.com/yt-dlp/yt-dlp/releases/tag/2021.09.02

Yeah they should be i just want to match your setup exactly.

itsteddyyo:
> > ...Done in [cd64147](https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/cd64147ad47bb80f6bcd1ceff262d932ff8e24a3).
> > Fixed for most characters. It currently does not work for special characters like the heart one in your example. Would that be good enough or should i keep looking?
> > Now completly fixed by switching to --dump-json in [f989da1](https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/f989da18c4c9256b8eb88a0bdd0cb228a3a941b0).
> 
> Regarding the search results, just wondering what are the benefits of searching with yt-dlp as opposed to how it was done before by manually parsing the results search page. I see that doing it this way is causing some issues, needs some extra steps like outputting to a local file, parsing, etc. and could be more error prone in certain circumstances, which makes me a little unsure if this is a needed or good change.
> 
What would those circumstances be? I get that the writing to file and parsing is a possible vulnerability theoretically but i didn't have issues with it so far. Did this cause problems for you? 😄 
> > The list from the manual downloading now gets populated by the result of yt-dlp too so the settings the user wants are represented here too. Plus i added the resolution in the list window.
> 
> I'm not sure what settings should be affecting the search, can you elaborate please?

The settings affecting the search are the filter/sort settings from yt-dlp. The videos are sorted by your preferences in "Extra Metadata Tools - Video - View Settings - Arguments for sorting videos" (maybe some user would prefer smaller filesize over quality). 
Plus videos not matching you settings get filtered out. You could for example say i dont want videos smaller than 1080p in "Arguments for filtering videos" and then i would not want to see those in the list as they could then not be downloaded later cause the downloading function uses the same filters and would just not download anything if your filter would not match the selected video.

darklinkpower:
> 
> 
> > > ...Done in [cd64147](https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/cd64147ad47bb80f6bcd1ceff262d932ff8e24a3).
> > > Fixed for most characters. It currently does not work for special characters like the heart one in your example. Would that be good enough or should i keep looking?
> > > Now completly fixed by switching to --dump-json in [f989da1](https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/f989da18c4c9256b8eb88a0bdd0cb228a3a941b0).
> > 
> > 
> > Regarding the search results, just wondering what are the benefits of searching with yt-dlp as opposed to how it was done before by manually parsing the results search page. I see that doing it this way is causing some issues, needs some extra steps like outputting to a local file, parsing, etc. and could be more error prone in certain circumstances, which makes me a little unsure if this is a needed or good change.
> 
> What would those circumstances be? I get that the writing to file and parsing is a possible vulnerability theoretically but i didn't have issues with it so far. Did this cause problems for you? 😄

Not about vulnerabilities, I was just thinking that the new approach to get results got a little more complex as opposed to the previous simple solution but I understand by your other comment that it needs to be done that way to filter the results. My take on this is that using a blackbox that we can't have control over and know what and how is doing something is something that I'm a little reluctant if it's something that can be done by yourself, knowing what and how things are being done. The error prone I mention is related to this, since being reliant on a 3rd party library or software for certain functionality can make it difficult to identify issues and solutions since you don't know what is going and you can only make guesses sometimes and it can also increase resources usage and add overhead in general. The issues in my other comments are examples of this: Issues with certain characters, returning only a few results for some reason, opening a window that you have to close manually, etc but it's good that you managed to resolve the issues.

darklinkpower:
I'm testing the extension with your latest changes:
- Settings:
image


- Results are taking a considerable time to be obtained: 67 seconds. Also the users can't know what is happening during this time

https://user-images.githubusercontent.com/1389286/134681307-a3e91c8d-1177-4c26-9e66-3937d633cf67.mp4

- Related to my previous comment and although maybe not by much, yt-dlp increases the network and memory usage by a certain amount 
image


I can only guess but I can think that it's taking that amount of time to get the results because it could be requesting the data of each video individually, getting their information and possibly filtering the results. This and what I've mentioned in my other comments make me not confident in making these changes unless absolutely necessary. I think there are two options:

- As the ideal solution, first investigate if we can do the filtering ourselves if the user has configured any filtering and completely remove the new dependency on yt-dlp to get results and only use it for video downloads.
- Use the previous method to get results when there are no filters configured or set to any and only use yt-dlp when there are filters, but also adding a notice in the settings window that filtering can produce a slowdown in getting results from the search.

darklinkpower:
I looked up a page source and made this regex to get the embedded json with the video info

`var ytInitialPlayerResponse = ((.*?(?=(;var meta))))`

itsteddyyo:
> I'm testing the extension with your latest changes:
> 
> * Settings:
>   image
> * Results are taking a considerable time to be obtained: 67 seconds. Also the users can't know what is happening during this time
> 
>  qfaAaojygC.mp4 
> * Related to my previous comment and although maybe not by much, yt-dlp increases the network and memory usage by a certain amount
>   image
> 
> I can only guess but I can think that it's taking that amount of time to get the results because it could be requesting the data of each video individually, getting their information and possibly filtering the results. This and what I've mentioned in my other comments make me not confident in making these changes unless absolutely necessary. I think there are two options:
> 
> * As the ideal solution, first investigate if we can do the filtering ourselves if the user has configured any filtering and completely remove the new dependency on yt-dlp to get results and only use it for video downloads.
> * Use the previous method to get results when there are no filters configured or set to any and only use yt-dlp when there are filters, but also adding a notice in the settings window that filtering can produce a slowdown in getting results from the search.

I also think yt-dlp is checking the video metadata one by one. 
Thanks I will look into the other options😄 May take a little time to check the alternatives as i want to see if there is a way to still do this with yt-dlp cause i think it won't be as easy to maintain with two ways of obtaining data. Plus it will be hard to check for every possible way to filter videos.

I will also look into giving the user more feedback.

Will let you know if i can solve the problems 😄

PR #61, opened 2021-09-27
Add Import of Origin licenses to DateImporter

itsteddyyo:
Added functionality for import Origin licenses. 

I oriented my code on what was used in the existing importers. The logic for retrieving the data from Origin was straight up stolen from Crow himself (https://github.com/JosefNemec/PlayniteExtensions/blob/9ec2f43a17f93707cfa39d6b4056a9459852a8fa/source/Libraries/OriginLibrary/Services/OriginAccountClient.cs) 😄 
When fetching the data from Origin only games and demos are considered, dlc or ingame purchases are filtered out before using your logic for processing the licenses.

I also added two new translation strings apart from the default ones for new importers:
- LOCDate_Importer_MenuItemImportOriginInvalidJson: for invalid json from Origin
- LOCDate_Importer_MenuItemImportOriginErrorMessage: for connection errors in the subsequent REST calls for retrieving the license data

There are apparently some games for which it will not work. For me it was Mass Effect 2/3 which have the release date included in the name in the Origin response so the name matching didn't work. I did fix this as it was only two games out of 45 for me and the user get the result export in the end anyway. I thought i would rather leave the code more simple than try to implement some complicated name matching. Hope thats ok 😄

### DIFF
--- a/source/Generic/DateImporter/DateImporter.psm1
+++ b/source/Generic/DateImporter/DateImporter.psm1
@@ -948,7 +948,7 @@ function Get-OriginLicenses
 
     $jsonValid = Get-IsJsonValidFromPage $pageSource
 
-    if($jsonValid -eq $false) 
+    if ($jsonValid -eq $false) 
     {
         $PlayniteApi.Dialogs.ShowMessage([Playnite.SDK.ResourceProvider]::GetString("LOCDate_Importer_MenuItemImportOriginInvalidJson"), "$libraryName Date Importer")
         return $LicensesList
@@ -989,7 +989,7 @@ function Get-OriginLicenses
     $licenses = Select-Xml $xml -XPath '/entitlements/entitlement'
     foreach ($license in $licenses) {
         $type = $license.Node.offer.offerType.ToLower() -replace " ", ""
-        if($($type -eq "basegame") -or $($type -eq "demo")) 
+        if ($($type -eq "basegame") -or $($type -eq "demo")) 
         {
             $date = Get-Date $license.Node.grantDate
             $product = [PSCustomObject]@{
--- a/source/Generic/DateImporter/DateImporter.psm1
+++ b/source/Generic/DateImporter/DateImporter.psm1
@@ -966,6 +966,8 @@ function Get-OriginLicenses
     try {
         $idResponse = Invoke-RestMethod -Method "GET" -Uri $idUrl -Headers $headers -ContentType "application/json" -UseBasicParsing
     } catch {
+        $errorMessage = $_.Exception.Message
+        $__logger.Info("Error retrieving user id from Origin. Uri: $idUrl,  Error: $errorMessage")
         $PlayniteApi.Dialogs.ShowMessage([Playnite.SDK.ResourceProvider]::GetString("LOCDate_Importer_MenuItemImportOriginErrorMessage"), "$libraryName Date Importer")
         return $LicensesList
     }
@@ -981,6 +983,8 @@ function Get-OriginLicenses
     try {
         $xml = Invoke-RestMethod -Method "GET" -Uri $licenseUrl -Headers $headers -ContentType "application/xml" -UseBasicParsing
     } catch {
+        $errorMessage = $_.Exception.Message
+        $__logger.Info("Error retrieving license data from Origin. Uri: $licenseUrl,  Error: $errorMessage")
         $PlayniteApi.Dialogs.ShowMessage([Playnite.SDK.ResourceProvider]::GetString("LOCDate_Importer_MenuItemImportOriginErrorMessage"), "$libraryName Date Importer")
         return $LicensesList
     }
--- a/source/Generic/DateImporter/DateImporter.psm1
+++ b/source/Generic/DateImporter/DateImporter.psm1
@@ -948,7 +948,8 @@ function Get-OriginLicenses
 
     $jsonValid = Get-IsJsonValidFromPage $pageSource
 
-    if($jsonValid -eq $false) {
+    if($jsonValid -eq $false) 
+    {
         $PlayniteApi.Dialogs.ShowMessage([Playnite.SDK.ResourceProvider]::GetString("LOCDate_Importer_MenuItemImportOriginInvalidJson"), "$libraryName Date Importer")
         return $LicensesList
     }
--- a/source/Generic/DateImporter/DateImporter.psm1
+++ b/source/Generic/DateImporter/DateImporter.psm1
@@ -964,7 +964,7 @@ function Get-OriginLicenses
     }
     $idResponse = $null
     try {
-        $idResponse = Invoke-RestMethod -Method "GET" -Uri $idUrl -Headers $headers -ContentType "application/json"
+        $idResponse = Invoke-RestMethod -Method "GET" -Uri $idUrl -Headers $headers -ContentType "application/json" -UseBasicParsing
     } catch {
         $PlayniteApi.Dialogs.ShowMessage([Playnite.SDK.ResourceProvider]::GetString("LOCDate_Importer_MenuItemImportOriginErrorMessage"), "$libraryName Date Importer")
         return $LicensesList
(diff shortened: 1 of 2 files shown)


source/Generic/DateImporter/DateImporter.psm1
```diff
@@ -846,5 +883,123 @@ function Get-SteamLicenses
...
+    }
+
+    $webView = $PlayniteApi.WebViews.CreateOffscreenView()
+    $webView.NavigateAndWait($authUrl)
+    $pageSource = $webView.GetPageSource()
+    $webView.Dispose()
+
+    $jsonValid = Get-IsJsonValidFromPage $pageSource
+
+    if($jsonValid -eq $false)
```
darklinkpower, line 951: can you add a space after the `if` here please?

source/Generic/DateImporter/Localization/en_US.xaml
```diff
@@ -11,7 +11,12 @@
     <sys:String x:Key="LOCDate_Importer_MenuItemImportSelectedEpicDatesDescription">Import dates of selected Epic games</sys:String>
     <sys:String x:Key="LOCDate_Importer_MenuItemImportAllEpicDatesDescription">Import dates of all Epic games</sys:String>
     <sys:String x:Key="LOCDate_Importer_MenuItemExportEpicLicencesDescription">Export Epic licenses list</sys:String>
-    <sys:String x:Key="LOCDate_Importer_LoginNotifyMessage">A web browser window will be opened, please close the window after login in.</sys:String>
+	<sys:String x:Key="LOCDate_Importer_MenuItemImportSelectedOriginDatesDescription">Import dates of selected Origin games</sys:String>
```
darklinkpower, line 14: The new loc strings are using tabs for indentations, can you change it to 4 spaces instead?

source/Generic/DateImporter/DateImporter.psm1
```diff
@@ -846,5 +883,123 @@ function Get-SteamLicenses
...
+    $token = $authJson.access_token
+
+
+    $idUrl = "https://gateway.ea.com/proxy/identity/pids/me"
+    $headers = @{
+        Authorization = "Bearer {0}" -f $token
+    }
+    $idResponse = $null
+    try {
+        $idResponse = Invoke-RestMethod -Method "GET" -Uri $idUrl -Headers $headers -ContentType "application/json"
```
darklinkpower, line 967: Can you please change the request to use a net class instead of this? For the reason that it will fail if internet explorer is not configured, see: https://stackoverflow.com/questions/38005341/the-response-content-cannot-be-parsed-because-the-internet-explorer-engine-is-no

Or maybe check if it works with `-UseBasicParsing` as mentioned there

source/Generic/DateImporter/DateImporter.psm1
```diff
@@ -846,5 +883,123 @@ function Get-SteamLicenses
...
+
+
+    $licenseTemplate = "https://api1.origin.com/ecommerce2/consolidatedentitlements/{0}?machine_hash=1"
+    $licenseUrl = $licenseTemplate -f $id 
+    $headers = @{
+        authtoken = $token
+    }
+    $xml = $null
+    try {
+        $xml = Invoke-RestMethod -Method "GET" -Uri $licenseUrl -Headers $headers -ContentType "application/xml"
```
darklinkpower, line 982: same here

darklinkpower (changes requested):

source/Generic/DateImporter/DateImporter.psm1
```diff
@@ -846,5 +883,123 @@ function Get-SteamLicenses
...
+    } catch {
+        $PlayniteApi.Dialogs.ShowMessage([Playnite.SDK.ResourceProvider]::GetString("LOCDate_Importer_MenuItemImportOriginErrorMessage"), "$libraryName Date Importer")
+        return $LicensesList
+    }
+
+
+    $licenses = Select-Xml $xml -XPath '/entitlements/entitlement'
+    foreach ($license in $licenses) {
+        $type = $license.Node.offer.offerType.ToLower() -replace " ", ""
+        if($($type -eq "basegame") -or $($type -eq "demo"))
```
darklinkpower, line 992: A space after the `if` here too please

source/Generic/DateImporter/DateImporter.psm1
itsteddyyo: Done in 559566c951bf00f312b54bf5f8efed31753055ee. 
Would you happen to have a config file for your IDE with your formatting preferences? So i don't need to get on your nerves with that stuff 😄

source/Generic/DateImporter/DateImporter.psm1
itsteddyyo: Done in 559566c951bf00f312b54bf5f8efed31753055ee.

source/Generic/DateImporter/DateImporter.psm1
itsteddyyo: Done in 85f6614eb5b9b7fa4801790c894fb5a2fd724f1f.
I added the -UseBasicParsing and it still works. Seems it's no longer neccessary for Powershell 6 and higher (https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-restmethod?view=powershell-7.1) but good call either way. It would be an easy enough way to support older installations. 
Can't really test it as funny enough I'm not even on Powershell 6 and i don't encounter the problem. I also have IE uninstalled and it still worked (with and without -UseBasicParsing) 😄

source/Generic/DateImporter/DateImporter.psm1
darklinkpower: I've not looked into settting a config file to be honest, I'll look in the weekend how to do it.

source/Generic/DateImporter/DateImporter.psm1
darklinkpower: I could reproduce it before I configured IE and had reports of several people having errors for that reason before so I've switched to using net class since then. I just tested it and there are no problems. If someone reports an issue then we can change it.

source/Generic/DateImporter/DateImporter.psm1
```diff
@@ -846,5 +883,123 @@ function Get-SteamLicenses
...
+    $licenseTemplate = "https://api1.origin.com/ecommerce2/consolidatedentitlements/{0}?machine_hash=1"
+    $licenseUrl = $licenseTemplate -f $id 
+    $headers = @{
+        authtoken = $token
+    }
+    $xml = $null
+    try {
+        $xml = Invoke-RestMethod -Method "GET" -Uri $licenseUrl -Headers $headers -ContentType "application/xml" -UseBasicParsing
+    } catch {
+        $PlayniteApi.Dialogs.ShowMessage([Playnite.SDK.ResourceProvider]::GetString("LOCDate_Importer_MenuItemImportOriginErrorMessage"), "$libraryName Date Importer")
```
darklinkpower, line 984: Can you add error logging in the catch block to detect what the issue was please? Maybe like 

```
$errorMessage = $_.Exception.Message
$__logger.Info("Error during x thing. Uri: $theUsedUri,  Error: $errorMessage
```

source/Generic/DateImporter/DateImporter.psm1
```diff
@@ -846,5 +883,123 @@ function Get-SteamLicenses
...
+
+    $idUrl = "https://gateway.ea.com/proxy/identity/pids/me"
+    $headers = @{
+        Authorization = "Bearer {0}" -f $token
+    }
+    $idResponse = $null
+    try {
+        $idResponse = Invoke-RestMethod -Method "GET" -Uri $idUrl -Headers $headers -ContentType "application/json" -UseBasicParsing
+    } catch {
+        $PlayniteApi.Dialogs.ShowMessage([Playnite.SDK.ResourceProvider]::GetString("LOCDate_Importer_MenuItemImportOriginErrorMessage"), "$libraryName Date Importer")
```
darklinkpower, line 969: Same here

darklinkpower (changes requested):

darklinkpower:
Also these strings need to be changed, adding `Origin`, in alphabetical order
image

itsteddyyo:
> Also these strings need to be changed, adding `Origin`, in alphabetical order
> image

Dammit i knew i forgot something when finishing the code. Just couldn't put my finger on it 😆
Done in 9b7b900e2bd5d779a5a5dd9b088920f0fa4201ed.

source/Generic/DateImporter/DateImporter.psm1
itsteddyyo: Done in 30f239ed4f0a4e4f132c2cadb9d3397dddc12f63.

darklinkpower (approved):

Issue #122, opened 2021-11-23
[PlayState] Add support for sharing real play time to another extensions


BanCrash:
**Extension**
PlayState

**Describe the solution you'd like**
Add support for sharing the real play time after using the pause feature to another extensions, like Game Activity.

Opening this to keep tracking here too.

Quote from @Lacro59 on https://github.com/Lacro59/playnite-gameactivity-plugin/issues/55 :

> @darklinkpower
> The ideal is for the plugin to share the different poses to remove them from the session.
> But it will probably be necessary to wait until Playnite.SDK allows to share the data.

darklinkpower:
Depends on https://github.com/JosefNemec/Playnite/issues/2039

PR #193, opened 2022-03-01
Added temporary workaround for sharing paused time

BanCrash:
I have verified that:

This PR is linked to this PR of GameActivity extension: https://github.com/Lacro59/playnite-gameactivity-plugin/pull/119

Temporary workaround for #122

I used the simple code as possible since this is only a temporary workaround until Playnite allows to share data among extensions.

The file is placed inside AppData/Roaming/Playnite/ExtensionsData folder with the name of "PlayState.txt". This file will be cleaned after executing one game or retrieving the info on GameActivity plugin in order to avoid reusing the same information over and over again if PlayState gets uninstalled or is not working for any reason.

I added as much code as possible in new methods, notes for explaining the functioning and a annotation that says:

> // Temporary workaround for sharing PlayState paused time until Playnite allows to share data among extensions

In order to ease the removal in the future when another better method is available.

EDIT: I forgot to mention, I tested the extension manually replacing the compiled files, I didn't achieve to create the pext file.

### DIFF
--- a/source/Generic/PlayState/PlayState.cs
+++ b/source/Generic/PlayState/PlayState.cs
@@ -599,7 +599,7 @@ namespace PlayState
             // This method will remove the info of the txt file in order to avoid reusing the previous play information.
             string[] info = { " ", " " };
 
-            File.WriteAllLines(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Playnite", "ExtensionsData", "PlayState.txt"), info);
+            File.WriteAllLines(Path.Combine(PlayniteApi.Paths.ExtensionsDataPath, "PlayState.txt"), info);
         }
 
         private void ExportPausedTimeInfo(Game game, ulong elapsedSeconds) // Temporary workaround for sharing PlayState paused time until Playnite allows to share data among extensions
@@ -607,7 +607,7 @@ namespace PlayState
             // This method will write the Id and pausedTime to PlayState.txt file placed inside ExtensionsData Roaming Playnite folder
             string[] info = { game.Id.ToString(), elapsedSeconds.ToString() };
 
-            File.WriteAllLines(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Playnite", "ExtensionsData", "PlayState.txt"), info);
+            File.WriteAllLines(Path.Combine(PlayniteApi.Paths.ExtensionsDataPath, "PlayState.txt"), info);
         }
 
         public override IEnumerable<MainMenuItem> GetMainMenuItems(GetMainMenuItemsArgs args)
--- a/source/Generic/PlayState/PlayState.cs
+++ b/source/Generic/PlayState/PlayState.cs
@@ -228,6 +228,8 @@ namespace PlayState
 
         public override void OnGameStarted(OnGameStartedEventArgs args)
         {
+            InitializePlaytimeInfoFile(); // Temporary workaround for sharing PlayState paused time until Playnite allows to share data among extensions
+
             if (isSuspended)
             {
                 SwitchGameState();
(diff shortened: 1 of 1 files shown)


source/Generic/PlayState/PlayState.cs
```diff
@@ -591,6 +594,22 @@ private void RemoveGameStopwatchTuple(Game game)
             }
         }
 
+        private void InitializePlaytimeInfoFile() // Temporary workaround for sharing PlayState paused time until Playnite allows to share data among extensions
+        {
+            // This method will remove the info of the txt file in order to avoid reusing the previous play information.
+            string[] info = { " ", " " };
+
+            File.WriteAllLines(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Playnite", "ExtensionsData", "PlayState.txt"), info);
```
darklinkpower, line 602: Please use the `ExtensionsDataPath` path from the Paths API and not environment variables https://playnite.link/docs/api/Playnite.SDK.IPlaynitePathsAPI.html#Playnite_SDK_IPlaynitePathsAPI_ExtensionsDataPath

Should be `PlayniteApi.Paths.ExtensionsDataPath`. You can use the `ConfigurationPath` if you prefer also but preferably `ExtensionsDataPath`

source/Generic/PlayState/PlayState.cs
```diff
@@ -591,6 +594,22 @@ private void RemoveGameStopwatchTuple(Game game)
...
+        }
+
+        private void ExportPausedTimeInfo(Game game, ulong elapsedSeconds) // Temporary workaround for sharing PlayState paused time until Playnite allows to share data among extensions
+        {
+            // This method will write the Id and pausedTime to PlayState.txt file placed inside ExtensionsData Roaming Playnite folder
+            string[] info = { game.Id.ToString(), elapsedSeconds.ToString() };
+
+            File.WriteAllLines(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Playnite", "ExtensionsData", "PlayState.txt"), info);
```
darklinkpower, line 610: Same here

darklinkpower (changes requested):

darklinkpower:
Another thing. Is the order of execution in the `OnGameStopped` PlayState -> GameActivity? Because if it's the other way around, then this will not work.

source/Generic/PlayState/PlayState.cs
BanCrash: I will check it, thanks!

BanCrash:
@darklinkpower done!

darklinkpower:
Be aware that `ExtensionsDataPath` is plugin specific so you will need to combine paths to get the location of this plugin's `ExtensionsDataPath` from another plugin.  `ConfigurationPath` is global for all plugins so it's also ok if you want to use that instead.

BanCrash:
> Be aware that `ExtensionsDataPath` is plugin specific so you will need to combine paths to get the location of this plugin's `ExtensionsDataPath` from another plugin. `ConfigurationPath` is global for all plugins so it's also ok if you want to use that instead.

I already tested with the last version of this PR and the last version of GameActivity PR and it's working as before changing the path. GameActivity is reading the file and substracting the paused time from its data as expected:

image
(After few seconds I paused for 1 minute and this is the result).

Also, I modified now the PlayState to log the path to assure it and this is the result:
image

> 01-03 23:28:38.147|DEBUG|PlayState#PlayState:C:\Users\User\AppData\Roaming\Playnite\ExtensionsData\PlayState.txt

darklinkpower:
Oh you are right, I was confusing it with this, my bad https://playnite.link/docs/api/Playnite.SDK.Plugins.Plugin.html#Playnite_SDK_Plugins_Plugin_GetPluginUserDataPath

PR #194, opened 2022-03-07
PlayState: Add Windows notifications

BanCrash:
I have verified that:

Changes:
- Added a setting for set a hotkey to show information about the current session of the game.
- Added a setting for use Windows notifications instead Playnite overlay system.
- Show information about play time on the notifications when Windows notifications are enabled.

As stated on the new settings note, in order to Windows notifications be showed it requires that Playnite is added to the priority list if the Focuss Assist is enabled when playing a game.

The notifications will be still showed after finishing the game since currentGame doesn't get deleted after finishing the game. I don't see any reason to not remove it on OnGameStopped event. Could you confirm that this is the case so I'll delete currentGame variable when finishing the game in order to avoid notifications about an ended game?

### DIFF
--- a/source/Generic/PlayState/Localization/en_US.xaml
+++ b/source/Generic/PlayState/Localization/en_US.xaml
@@ -3,8 +3,6 @@
     <sys:String x:Key="LOCPlayState_MenuItemRemoveFromBlacklistDescription">Remove selected games from Playstate Blacklist</sys:String>
     <sys:String x:Key="LOCPlayState_MenuItemAddToPlaytimeSuspendDescription">Configure selected games to force to only suspend playtime</sys:String>
     <sys:String x:Key="LOCPlayState_MenuItemRemoveFromPlaytimeSuspendDescription">Remove configuration to force to only suspend playtime</sys:String>
-    <sys:String x:Key="LOCPlayState_MenuItemAddToWindowsNotificationsStyleDescription">Configure selected games to force to show Windows notifications</sys:String>
-    <sys:String x:Key="LOCPlayState_MenuItemRemoveFromWindowsNotificationsStyleDescription">Remove configuration to force to show Windows notifications</sys:String>
     <sys:String x:Key="LOCPlayState_NotificationMessageHotkeyRegisterFailed" xml:space="preserve">Failed to register configured Hotkey {0}
 
 You can try using a diferent HotKey combination to possibly fix the issue</sys:String>
@@ -14,10 +12,6 @@ You can try using a diferent HotKey combination to possibly fix the issue</sys:S
     <sys:String x:Key="LOCPlayState_PlaytimeSuspendRemovedResultsMessage" xml:space="preserve">Removed configuration to force to only suspend playtime when using the configured hotkey from {0} game(s).
 
 Games will be suspended when using the configured hotkey.</sys:String>
-    <sys:String x:Key="LOCPlayState_WindowsNotificationsStyleAddedResultsMessage">Configured {0} game(s) to force to show Windows notifications instead Playnite overlay.</sys:String>
-    <sys:String x:Key="LOCPlayState_WindowsNotificationsStyleRemovedResultsMessage" xml:space="preserve">Removed configuration to force to show Windows notifications from {0} game(s).
-
-Notifications will be send using Playnite overlay.</sys:String>
     <sys:String x:Key="LOCPlayState_StatusSuspendedMessage">Game suspended</sys:String>
     <sys:String x:Key="LOCPlayState_StatusResumedMessage">Game resumed</sys:String>
     <sys:String x:Key="LOCPlayState_StatusPlaytimeSuspendedMessage">Playtime suspended</sys:String>
@@ -39,9 +33,7 @@ If you have Focus Assist enabled when playing a game, you have to add Playnite t
     
 This can also be applied individually per game by adding a feature named "[PlayState] Suspend playtime only" to a game</sys:String>
     <sys:String x:Key="LOCPlayState_SettingGlobalShowWindowsNotificationsStyleLabel">Show Windows notifications instead Playnite overlay</sys:String>
-    <sys:String x:Key="LOCPlayState_SettingGlobalShowWindowsNotificationsStyleTooltip" xml:space="preserve">By enabling this option, all games will show Windows toast notifications instead Playnite overlay notifications.
-    
-This can also be applied individually per game by adding a feature named "[PlayState] Show Windows notifications style" to a game</sys:String>
+    <sys:String x:Key="LOCPlayState_SettingGlobalShowWindowsNotificationsStyleTooltip" xml:space="preserve">By enabling this option, games will show Windows toast notifications instead Playnite overlay notifications.</sys:String>
     <sys:String x:Key="LOCPlayState_Playtime">Playtime:</sys:String>
     <sys:String x:Key="LOCPlayState_TotalPlaytime">Total playtime:</sys:String>
     <sys:String x:Key="LOCPlayState_HoursMinutesPlayed">{0} hours {1} minutes</sys:String>
--- a/source/Generic/PlayState/Localization/en_US.xaml
+++ b/source/Generic/PlayState/Localization/en_US.xaml
@@ -32,8 +32,8 @@ If you have Focus Assist enabled when playing a game, you have to add Playnite t
     <sys:String x:Key="LOCPlayState_SettingGlobalOnlySuspendPlaytimeTooltip" xml:space="preserve">By enabling this option, all games won't be suspended when using the configured hotkey and only the time will be counted to substracted from the game playtime when it stops.
     
 This can also be applied individually per game by adding a feature named "[PlayState] Suspend playtime only" to a game</sys:String>
-    <sys:String x:Key="LOCPlayState_SettingGlobalShowWindowsNotificationsStyleLabel">Show Windows notifications instead Playnite overlay</sys:String>
-    <sys:String x:Key="LOCPlayState_SettingGlobalShowWindowsNotificationsStyleTooltip" xml:space="preserve">By enabling this option, games will show Windows toast notifications instead Playnite overlay notifications.</sys:String>
+    <sys:String x:Key="LOCPlayState_SettingGlobalShowWindowsNotificationsStyleLabel">Show native Windows notifications instead of Playnite notification window</sys:String>
+    <sys:String x:Key="LOCPlayState_SettingGlobalShowWindowsNotificationsStyleTooltip">By enabling this option, games will show native Windows notifications instead of Playnite notification window.</sys:String>
     <sys:String x:Key="LOCPlayState_Playtime">Playtime:</sys:String>
     <sys:String x:Key="LOCPlayState_TotalPlaytime">Total playtime:</sys:String>
     <sys:String x:Key="LOCPlayState_HoursMinutesPlayed">{0} hours {1} minutes</sys:String>
--- a/source/Generic/PlayState/Localization/en_US.xaml
+++ b/source/Generic/PlayState/Localization/en_US.xaml
@@ -3,6 +3,8 @@
     <sys:String x:Key="LOCPlayState_MenuItemRemoveFromBlacklistDescription">Remove selected games from Playstate Blacklist</sys:String>
     <sys:String x:Key="LOCPlayState_MenuItemAddToPlaytimeSuspendDescription">Configure selected games to force to only suspend playtime</sys:String>
     <sys:String x:Key="LOCPlayState_MenuItemRemoveFromPlaytimeSuspendDescription">Remove configuration to force to only suspend playtime</sys:String>
+    <sys:String x:Key="LOCPlayState_MenuItemAddToWindowsNotificationsStyleDescription">Configure selected games to force to show Windows notifications</sys:String>
+    <sys:String x:Key="LOCPlayState_MenuItemRemoveFromWindowsNotificationsStyleDescription">Remove configuration to force to show Windows notifications</sys:String>
     <sys:String x:Key="LOCPlayState_NotificationMessageHotkeyRegisterFailed" xml:space="preserve">Failed to register configured Hotkey {0}
 
 You can try using a diferent HotKey combination to possibly fix the issue</sys:String>
@@ -12,17 +14,23 @@ You can try using a diferent HotKey combination to possibly fix the issue</sys:S
     <sys:String x:Key="LOCPlayState_PlaytimeSuspendRemovedResultsMessage" xml:space="preserve">Removed configuration to force to only suspend playtime when using the configured hotkey from {0} game(s).
 
 Games will be suspended when using the configured hotkey.</sys:String>
+    <sys:String x:Key="LOCPlayState_WindowsNotificationsStyleAddedResultsMessage">Configured {0} game(s) to force to show Windows notifications instead Playnite overlay.</sys:String>
+    <sys:String x:Key="LOCPlayState_WindowsNotificationsStyleRemovedResultsMessage" xml:space="preserve">Removed configuration to force to show Windows notifications from {0} game(s).
+
+Notifications will be send using Playnite overlay.</sys:String>
     <sys:String x:Key="LOCPlayState_StatusSuspendedMessage">Game suspended</sys:String>
     <sys:String x:Key="LOCPlayState_StatusResumedMessage">Game resumed</sys:String>
     <sys:String x:Key="LOCPlayState_StatusPlaytimeSuspendedMessage">Playtime suspended</sys:String>
     <sys:String x:Key="LOCPlayState_StatusPlaytimeResumedMessage">Playtime resumed</sys:String>
     <sys:String x:Key="LOCPlayState_SettingConfiguredHotkeyLabel">Configured Suspend/Resume hotkey:</sys:String>
+    <sys:String x:Key="LOCPlayState_SettingConfiguredInformationHotkeyLabel">Configured Information hotkey:</sys:String>
     <sys:String x:Key="LOCPlayState_SettingChangeHotkeyButtonLabel">Change Hotkey</sys:String>
     <sys:String x:Key="LOCPlayState_SettingPressHotkeyPromtButtonLabel">Press Hotkey now</sys:String>
     <sys:String x:Key="LOCPlayState_SettingNotes" xml:space="preserve">Notes:
 
 Configured hotkey won't be able to be used in other applications.
-Changing this setting requires restart.</sys:String>
+Changing this setting requires restart.
+If you have Focus Assist enabled when playing a game, you have to add Playnite to priority list in order to receive Windows notifications.</sys:String>
     <sys:String x:Key="LOCPlayState_SettingSubstractSuspendedPlaytimeOnStoppedLabel">Substract the time the game was suspended from the game playtime when it is stopped</sys:String>
     <sys:String x:Key="LOCPlayState_SettingSubstractSuspendedPlaytimeOnStoppedTooltip">Helps to keep a more accurate playtime value in games by only keeping the time the game was not suspended count towards the game playtime</sys:String>
     <sys:String x:Key="LOCPlayState_SettingSubstractOnlyNonLibraryGamesLabel">Only execute for games that are not managed by library plugins</sys:String>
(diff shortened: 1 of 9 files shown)


source/Generic/PlayState/PlayState.cs
```diff
@@ -242,6 +267,14 @@ public override void OnGameStarted(OnGameStartedEventArgs args)
                 return;
             }
 
+            playedDateTime = DateTime.Now;
+
+            showWindowsNotificationsStyle = false;
+            if (settings.Settings.GlobalShowWindowsNotificationsStyle)
+            {
+                showWindowsNotificationsStyle = true;
+            }
```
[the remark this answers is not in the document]
darklinkpower: A class could be created that has the info used by the extension, maybe like this

```csharp
    class PlaystateData
    {
        public id DatabaseId { get; set; }
        public datetime StartDate { get; set; }
        public Stopwatch Stopwatch { get; set; }
    }
```

This could replace the `List<Tuple<Guid, Stopwatch>>` list and use a `List<PlaystateData>` list instead to use in the extension. Maybe even put the revelant methods inside the class, like the ones to get the current playtime or so, but that's just an idea.

source/Generic/PlayState/PlayState.cs
```diff
@@ -392,6 +425,175 @@ private void WindowClosed(object sender, EventArgs e)
...
+        /// - "suspended" for suspend process and playtime<br/>
+        /// - "playtimeSuspended" for suspend playtime<br/>
+        /// - "information" for showing the actual status<br/>
+        /// </param>
+        /// </summary>
+        private void ShowNotification(string status)
+        {
+            switch (status)
+            {
+                case "resumed": // for resuming process and playtime
```
darklinkpower, line 499: Change to enums please

source/Generic/PlayState/PlayState.cs
```diff
@@ -392,6 +425,175 @@ private void WindowClosed(object sender, EventArgs e)
...
+                    }
+                    else
+                    {
+                        // Windows notification only admits 3 .Addtext properties, so for more text lines you have to use {Environment.NewLine}
+                        new ToastContentBuilder()
+                            .AddText(currentGame.Name) // First AddText field will act as a title
+                            .AddText(ResourceProvider.GetString("LOCPlayState_StatusResumedMessage"))
+                            .AddText($"{ResourceProvider.GetString("LOCPlayState_Playtime")} {GetHoursString(GetRealPlaytime())}" +
+                                $"{Environment.NewLine}{ResourceProvider.GetString("LOCPlayState_TotalPlaytime")} {GetHoursString(GetRealPlaytime() + currentGame.Playtime)}")
+                            .AddHeroImage(new Uri(Path.Combine(PlayniteApi.Paths.ConfigurationPath, "library/files", currentGame.BackgroundImage))) // Show game image in the notification
```
darklinkpower, line 512: Don't combine the path here. Use this method instead https://playnite.link/docs/master/api/Playnite.SDK.IGameDatabaseAPI.html#Playnite_SDK_IGameDatabaseAPI_GetFullFilePath_System_String_

Example: https://playnite.link/docs/master/tutorials/extensions/library.html#exporting-game-cover

Also, are the toasts compatible with urls? In case it doesn't, then that needs to be handled in case the game has an url set for the image

source/Generic/PlayState/PlayState.cs
```diff
@@ -429,11 +631,11 @@ private void SwitchGameState()
                         isSuspended = false;
                         if (processesSuspended)
                         {
-                            ShowSplashWindow(ResourceProvider.GetString("LOCPlayState_StatusResumedMessage"));
+                            ShowNotification("resumed");
```
darklinkpower, line 634: Enums

source/Generic/PlayState/PlayState.csproj
```diff
@@ -12,6 +12,8 @@
     <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <Deterministic>true</Deterministic>
+    <NuGetPackageImportStamp>
```
darklinkpower, line 15: What is this?

source/Generic/PlayState/PlayState.csproj
```diff
@@ -31,9 +33,6 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Playnite.SDK, Version=6.2.2.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>packages\PlayniteSDK.6.2.2\lib\net462\Playnite.SDK.dll</HintPath>
-    </Reference>
```
darklinkpower, line 36: Why was this removed?

source/Generic/PlayState/PlayState.csproj
```diff
@@ -68,7 +67,6 @@
     <None Include="extension.yaml">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Include="packages.config" />
```
darklinkpower, line 71: Why was this removed?

source/Generic/PlayState/PlayState.csproj
```diff
@@ -93,6 +91,17 @@
...
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Toolkit.Uwp.Notifications">
+      <Version>7.1.2</Version>
+    </PackageReference>
+    <PackageReference Include="PlayniteSDK">
+      <Version>6.2.2</Version>
+    </PackageReference>
+    <PackageReference Include="System.Management">
+      <Version>5.0.0</Version>
+    </PackageReference>
```
darklinkpower, line 103: Is this reference used in the project?

source/Generic/PlayState/PlayStateSettings.cs
```diff
@@ -56,12 +64,26 @@ public PlayStateSettingsViewModel(PlayState plugin)
...
+                    
+                }
+                if (Settings.SavedInformationHotkeyGesture.Modifiers == ModifierKeys.None)
+                {
+                    // Due to a bug in previous version that allowed 
+                    // to save gestures without modifiers, this
+                    // should be done to restore the default ModifierKeys
+                    Settings.SavedInformationHotkeyGesture.Modifiers = (ModifierKeys)5;
+                    modifiedSettings = true;
+                }
```
darklinkpower, line 84: Why is this being done for the new hotkeys? The bug should not exist there right?

source/Generic/PlayState/packages.config
```diff
@@ -1,5 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<packages>
-  <package id="PlayniteSDK" version="6.2.2" targetFramework="net462" />
-  <package id="System.Management" version="5.0.0" targetFramework="net462" />
-</packages>
```
darklinkpower, line 5: Why was this removed?

darklinkpower (changes requested):

source/Generic/PlayState/Localization/en_US.xaml
```diff
@@ -17,17 +17,29 @@ Games will be suspended when using the configured hotkey.</sys:String>
...
     <sys:String x:Key="LOCPlayState_SettingSubstractSuspendedPlaytimeOnStoppedTooltip">Helps to keep a more accurate playtime value in games by only keeping the time the game was not suspended count towards the game playtime</sys:String>
     <sys:String x:Key="LOCPlayState_SettingSubstractOnlyNonLibraryGamesLabel">Only execute for games that are not managed by library plugins</sys:String>
     <sys:String x:Key="LOCPlayState_SettingGlobalOnlySuspendPlaytimeLabel">Only suspend playtime count globally and not the game processes</sys:String>
     <sys:String x:Key="LOCPlayState_SettingGlobalOnlySuspendPlaytimeTooltip" xml:space="preserve">By enabling this option, all games won't be suspended when using the configured hotkey and only the time will be counted to substracted from the game playtime when it stops.
     
 This can also be applied individually per game by adding a feature named "[PlayState] Suspend playtime only" to a game</sys:String>
+    <sys:String x:Key="LOCPlayState_SettingGlobalShowWindowsNotificationsStyleLabel">Show Windows notifications instead Playnite overlay</sys:String>
```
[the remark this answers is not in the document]
darklinkpower: I also think it should be changed "Playnite overlay" -> "Playnite notification window" since it's not an overlay. If you like and agree, perhaps changing it to 

> Show native Windows notifications instead of Playnite notification window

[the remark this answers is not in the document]
BanCrash: I like your change yeah, I will replace it.

source/Generic/PlayState/PlayState.cs
```diff
@@ -100,21 +104,38 @@ internal bool RegisterGlobalHotkey()
...
                     PlayniteApi.Notifications.Add(new NotificationMessage(Guid.NewGuid().ToString(),
                         "PlayState: " + string.Format(ResourceProvider.GetString("LOCPlayState_NotificationMessageHotkeyRegisterFailed"), settings.Settings.SavedHotkeyGesture),
                         NotificationType.Error));
-                    logger.Error($"Failed to register configured Hotkey {settings.Settings.SavedHotkeyGesture}.");
+                    logger.Error($"Failed to register configured pause/resume Hotkey {settings.Settings.SavedHotkeyGesture}.");
                 }
 
-                return registered;
+                // Information Hotkey
+                var registered2 = HotkeyHelper.RegisterHotKey(handle, HOTKEY_ID, settings.Settings.SavedInformationHotkeyGesture.Modifiers.ToVK(), (uint)KeyInterop.VirtualKeyFromKey(settings.Settings.SavedInformationHotkeyGesture.Key));
```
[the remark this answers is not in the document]
BanCrash: The new hotkey is for just showing information about the actual state, not changes the state of paused/resumed. 

About the second, I didn't understand well the logic behind the hotkey system tbh, but in my tests it's working as expected.

source/Generic/PlayState/PlayState.cs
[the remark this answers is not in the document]
BanCrash: I just followed the code as paused/resumed setting was being handled there too. But you are right, should be on the ShowNotification method.

I will see if I can make work the second comment.

source/Generic/PlayState/PlayState.cs
BanCrash: I checked but didn't find a better way, thanks for pointing me out.

The toasts are compatible with links, in their doc they have this example:
`.AddInlineImage(new Uri("https://picsum.photos/360/202?image=883"))`

https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/send-local-toast?tabs=uwp

source/Generic/PlayState/PlayState.csproj
BanCrash: Aarrg, a file that showed the migration from packages.config to ReferencePackage. I removed the file but forgot about the reference here.

source/Generic/PlayState/PlayState.csproj
BanCrash: Was automatically removed by Visual Studio when changing to package reference: https://docs.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference

This change was necessary because it didn't compile after adding the notifications dependencies without doing so.

source/Generic/PlayState/PlayState.csproj
BanCrash: Was automatically removed by Visual Studio when changing to package reference: https://docs.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference

This change was necessary because it didn't compile after adding the notifications dependencies without doing so.

source/Generic/PlayState/PlayState.csproj
BanCrash: It was added automatically when adding the dependencies for Windows notifications, so I assume yes. Anyway I will make a test without it just to make sure.

source/Generic/PlayState/PlayStateSettings.cs
BanCrash: Like I said before I don't understand well the logic behind hotkeys. I thought this was executed after editing hotkeys all the time. I will remove it then.

source/Generic/PlayState/packages.config
BanCrash: Was automatically removed by Visual Studio when changing to package reference: https://docs.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference

This change was necessary because it didn't compile after adding the notifications dependencies without doing so.

BanCrash: Okay, seems that it wasn't necessary. I reverted this change and added notifications on packages.config file.

source/Generic/PlayState/PlayState.cs
```diff
@@ -134,6 +155,10 @@ private IntPtr GlobalHotkeyCallback(IntPtr hWnd, int msg, IntPtr wParam, IntPtr
                         {
                             SwitchGameState();
                         }
+                        else if (vkey == (uint)KeyInterop.VirtualKeyFromKey(settings.Settings.SavedInformationHotkeyGesture.Key))
+                        {
+                            ShowNotification("information");
```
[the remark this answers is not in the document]
BanCrash: Done

source/Generic/PlayState/PlayState.cs
BanCrash: Done

source/Generic/PlayState/PlayState.cs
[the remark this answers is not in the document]
BanCrash: I checked again. Like I've said I don't know the logic behind, but in my tests this is working as expected. If I use the hotkey for Pause/Resume, it does that, and if I use the hotkey for information, it doesn't resume but only show the information.

source/Generic/PlayState/PlayState.cs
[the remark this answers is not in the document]
BanCrash: Done

BanCrash:
I think I fixed all the issues @darklinkpower 
I create the new class as you suggested, and now when currentGame is finished, it will set currentGame to the last executed game if more than one were executing at the same time.

darklinkpower:
I wanted to build it before giving the last check but the Microsoft.Toolkit namespace was not being detected. Upon uninstalling the package and installing again, I get this

image

---

Another thing, here it says it's only for Windows 10. Can you add a check and only allow/try to send toasts if the user is on Windows 10? Could be a value set in the constructor

image

source/Generic/PlayState/PlayState.cs
```diff
@@ -494,6 +488,12 @@ private string GetHoursString(ulong playtimeSeconds)
         /// </summary>
         private void ShowNotification(string status)
         {
+            showWindowsNotificationsStyle = false;
```
darklinkpower, line 489: can be simplified as

```csharp
showWindowsNotificationsStyle = settings.Settings.GlobalShowWindowsNotificationsStyle
```

Also, is it needed to store the value in the plugin itself? Or can it be a variable inside this method like

```csharp
var showWindowsNotificationsStyle = settings.Settings.GlobalShowWindowsNotificationsStyle
```

source/Generic/PlayState/PlayState.cs
```diff
@@ -771,29 +773,32 @@ public override void OnGameStopped(OnGameStoppedEventArgs args)
...
+            if (playStateData.Any(x => x.Game.Id == game.Id && x.Stopwatch != null))
             {
-                logger.Debug($"A stopwatch for game {game.Name} with id {game.Id} already exists");
+                logger.Debug($"Data for game {game.Name} with id {game.Id} already exists");
             }
             else
             {
-                stopwatchList.Add(new Tuple<Guid, Stopwatch>(game.Id, new Stopwatch()));
-                logger.Debug($"Stopwatch for game {game.Name} with id {game.Id} was created");
+                playStateData.Add(new PlayStateData(game, DateTime.Now, new Stopwatch()));
```
darklinkpower, line 787: Do you think it would make sense to get the `DateTime.Now` value and create the stopwatch inside the class constructor? I think it would make more sense since this is something that doesn't need outside reference and will simplify the object creation

source/Generic/PlayState/PlayState.cs
```diff
@@ -771,29 +773,32 @@ public override void OnGameStopped(OnGameStoppedEventArgs args)
...
+                logger.Debug($"Data for game {game.Name} with id {game.Id} was created");
             }
         }
 
-        private void RemoveGameStopwatchTuple(Game game)
+        private void RemoveGame(Game game)
         {
-            var tuple = stopwatchList.FirstOrDefault(x => x.Item1 == game.Id);
-            if (tuple != null)
+            if (playStateData.Any(x => x.Game.Id == game.Id))
```
darklinkpower, line 794: Could use

```csharp
var gameData = FirstOrDefault(x => x.Game.Id == game.Id);
if (gameData != null)
{
    playStateData.Remove(gameData);
```

To prevent iterating twice

darklinkpower (changes requested):

BanCrash:
> I wanted to build it before giving the last check but the Microsoft.Toolkit namespace was not being detected. Upon uninstalling the package and installing again, I get this
> 
> image
> 
> Another thing, here it says it's only for Windows 10. Can you add a check and only allow/try to send toasts if the user is on Windows 10? Could be a value set in the constructor
> 
> image

Yes, that's the issue that force me to do the first change about removing packages.config and so. But the funny thing is now it's working for me. I will do the switch again then.

Yes, you are right, only wor W10. I will check it.

source/Generic/PlayState/PlayState.cs
```diff
@@ -392,16 +417,196 @@ private void WindowClosed(object sender, EventArgs e)
...
+        /// </summary>
+        private ulong GetRealPlaytime()
+        {
+            var suspendedTime = playStateData.FirstOrDefault(x => x.Game.Id == currentGame.Id)?.Stopwatch.Elapsed;
+            ulong elapsedSeconds = 0;
+            if (suspendedTime != null)
+            {
+                elapsedSeconds = Convert.ToUInt64(suspendedTime.Value.TotalSeconds);
+            }
+            return Convert.ToUInt64(DateTime.Now.Subtract(playStateData.FirstOrDefault(x => x.Game.Id == currentGame.Id).StartDate).TotalSeconds) - elapsedSeconds;
```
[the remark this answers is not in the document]
BanCrash: Shouldn't be the case since the comprobations should be made before, so yeah, I will go with First instead FirstOrDefault

source/Generic/PlayState/PlayState.cs
BanCrash: I did that way for following the previous code, but the second case you said I think it's better. I will use it.

source/Generic/PlayState/PlayState.cs
BanCrash: Shouldn't be any issue doing so, I will test it.

BanCrash:
@darklinkpower all changes done. I restore the switch to PackageReference, should compile for you now. 
For W10 or W11 comprobation I had to check registry since for some reason Environment.OSVersion is not working.

source/Generic/PlayState/PlayStateSettingsView.xaml.cs
```diff
@@ -149,5 +154,11 @@ private void SetInformationHotkey(KeyEventArgs e, TextBox tbInformationHotkey, B
             setInformationHotkeyButton.Content = ResourceProvider.GetString("LOCPlayState_SettingChangeHotkeyButtonLabel");
         }
 
+        private bool IsWindows10Or11()
+        {
+            var productName = (string) Microsoft.Win32.Registry.LocalMachine.OpenSubKey(@"Software\Microsoft\Windows NT\CurrentVersion")?.GetValue("productName");
```
darklinkpower, line 159: The key needs to be disposed by using Dispose() method or the `using (...` similar to here https://github.com/darklinkpower/PlayniteExtensionsCollection/blob/24798e311c1db8dfd8eebbd5d1240ad842b20a28/source/Generic/SpecialKHelper/SpecialKHelper.cs#L194-L212 but I will take care of it after merging

darklinkpower:
This should also go here since it's about that specific option but I'll take care of this 
image

darklinkpower:
And there should be margin here. I'll take care of it 
image

darklinkpower:
The message should be different when the action was pause/unpause game or is just displaying information since currently it's not clear if something was done

image

darklinkpower:
I'll merge the PR, I'll make some modifications to some things I think need small changes and release afterwards. Thanks for contributing!

BanCrash:
> I'll merge the PR, I'll make some modifications to some things I think need small changes and release afterwards. Thanks for contributing!

Sure, no problem, thanks!

source/Generic/PlayState/PlayStateSettingsView.xaml.cs
```diff
@@ -25,6 +26,10 @@ public PlayStateSettingsView()
             InitializeComponent();
             TbHotkey.PreviewKeyDown += HotkeyTextBox_PreviewKeyDown;
             TbInformationHotkey.PreviewKeyDown += InformationHotkeyTextBox_PreviewKeyDown;
+            if (!IsWindows10Or11())
```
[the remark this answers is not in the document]
darklinkpower: Got it. I changed it in https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/05da1f1d947b1dcba5ae5b08461928dcdb2b8abb to make it usable during the evaluation to select the notification style

And in here I disabled the option in settings with MVVM pattern which is preferable https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/804a886d5cfb62897e05fe4f9082e6c08822b432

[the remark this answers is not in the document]
BanCrash: Sure, perfect then!

PR #209, opened 2022-03-14
Improved processes handling

BanCrash:
- Added processes to PlayStateData, and their states (isSuspended, processesSuspended)
- When you have more than one game opened, now you will be able to pause/resume the currentGame always. So, if you open Game1, and after you open Game2, but after a while you close Game2, you will be able to fully control again Game1.
- Refactored code to allow in the future handle several games at the same time (show notifications for another game that is not the currentGame, pause/resume other game that is not the currentGame, etc).

I have verified that:

### DIFF
--- a/source/Generic/PlayState/PlayState.cs
+++ b/source/Generic/PlayState/PlayState.cs
@@ -545,7 +545,7 @@ namespace PlayState
             {
                 return;
             }
-            
+
             var sb = new StringBuilder();
             switch (status)
             {
--- a/source/Generic/PlayState/PlayState.cs
+++ b/source/Generic/PlayState/PlayState.cs
@@ -186,7 +186,11 @@ namespace PlayState
                         uint vkey = ((uint)lParam >> 16) & 0xFFFF;
                         if (vkey == (uint)KeyInterop.VirtualKeyFromKey(settings.Settings.SavedHotkeyGesture.Key))
                         {
-                            SwitchGameState(currentGame);
+                            var gameData = GetGameData(currentGame);
+                            if (gameData != null)
+                            {
+                                SwitchGameState(gameData);
+                            }
                         }
                         else if (vkey == (uint)KeyInterop.VirtualKeyFromKey(settings.Settings.SavedInformationHotkeyGesture.Key))
                         {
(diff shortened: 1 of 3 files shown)


source/Generic/PlayState/PlayState.cs
```diff
@@ -413,6 +411,21 @@ private bool CurrentGameChanged(Game game)
...
+        private PlayStateData GetGameData(Game game)
+        {
+            if (game == null || !playStateData.Any(x => x.Game.Id == game.Id))
+            {
+                return null;
+            }
+            else
+            {
+                return playStateData.Find(x => x.Game.Id == game.Id);
+            }
```
darklinkpower, line 426: I think this could be simplified by just using:

```csharp
return playStateData.FirstOrDefault(x => x.Game.Id == game.Id)
```

Currently it's enumeration twice, one for the `Any` and then in the `Find`. With what I propose it would just search once and if it doesn't find anything, it will return `null` anyway

source/Generic/PlayState/PlayState.cs
```diff
@@ -576,52 +596,48 @@ private void ShowNotification(NotificationTypes status)
 
             if (settings.Settings.NotificationShowSessionPlaytime)
             {
-                sb.Append($"\n{ResourceProvider.GetString("LOCPlayState_Playtime")} {GetHoursString(GetRealPlaytime())}");
+                sb.Append($"\n{ResourceProvider.GetString("LOCPlayState_Playtime")} {GetHoursString(GetRealPlaytime(game))}");
```
darklinkpower, line 599: In here, it executes `GetRealPlaytime` method. This method then does
```csharp
            var gameData = GetGameData(game);
            if (gameData == null)
            {
                return 0;
            }
```
This method then gets called again for `if (settings.Settings.NotificationShowTotalPlaytime)`
You are already checking if the game has data at the start of the method, I think you should just pass that same data as argument to the required methods

darklinkpower: To prevent searching for this same data multiple times

darklinkpower: I see that it's also searched for in the `SwitchGameState()` method. If possible, please make it so it only searches once for it

darklinkpower (changes requested):

BanCrash:
@darklinkpower all requested changes done I think.

PR #221, opened 2022-03-26
(PlayState) Added override suspend playtime / Set suspendPlaytimeOnly variable by game instead of globally

BanCrash:
I have verified that:

Divided in 2 commits to ease the review.

1st commit (Added override suspend playtime only setting):
- Replaced "[PlayState] Suspend playtime only" feature with "[PlayState] Override suspend playtime only setting"
This will make that you can enable the option of "Only suspend playtime count globally and not the game processes" and select a few games to pause playtime and game processes, and the opposite, disable that option and select a few games to only pause playtime.
- Migrate the previous feature to the new, and add a message to the user so they know that the new feature behave on a different way than the previous one.

image

2nd commit (Set suspendPlaytimeOnly variable in PlayStateData):
- Set the variable "suspendPlaytimeOnly" by game and not globally, since some games could have it enabled and others disabled. This fixes a bug that caused that when you launched a game that have it enabled, and after that launched a game that have it disabled, when closing the second game the first game couldn't be paused anymore.

### DIFF
--- a/source/Generic/PlayState/Localization/en_US.xaml
+++ b/source/Generic/PlayState/Localization/en_US.xaml
@@ -2,17 +2,17 @@
     <sys:String x:Key="LOCPlayState_SettingsHelpLabel">Help</sys:String>
     <sys:String x:Key="LOCPlayState_MenuItemAddToBlacklistDescription">Add selected games to PlayState Blacklist</sys:String>
     <sys:String x:Key="LOCPlayState_MenuItemRemoveFromBlacklistDescription">Remove selected games from Playstate Blacklist</sys:String>
-    <sys:String x:Key="LOCPlayState_MenuItemAddToPlaytimeSuspendDescription">Configure selected games to force to only suspend playtime</sys:String>
-    <sys:String x:Key="LOCPlayState_MenuItemRemoveFromPlaytimeSuspendDescription">Remove configuration to force to only suspend playtime</sys:String>
+    <sys:String x:Key="LOCPlayState_MenuItemAddToPlaytimeSuspendDescription">Configure selected games to override suspend playtime only setting</sys:String>
+    <sys:String x:Key="LOCPlayState_MenuItemRemoveFromPlaytimeSuspendDescription">Remove configuration to override suspend playtime only setting</sys:String>
     <sys:String x:Key="LOCPlayState_NotificationMessageHotkeyRegisterFailed" xml:space="preserve">Failed to register configured Hotkey {0}
 
 You can try using a diferent HotKey combination to possibly fix the issue</sys:String>
     <sys:String x:Key="LOCPlayState_BlacklistAddedResultsMessage">Added {0} game(s) to the PlayState blacklist.</sys:String>
     <sys:String x:Key="LOCPlayState_BlacklistRemovedResultsMessage">Removed {0} game(s) from the PlayState blacklist.</sys:String>
-    <sys:String x:Key="LOCPlayState_PlaytimeSuspendAddedResultsMessage">Configured {0} game(s) to force to only suspend playtime when using the configured hotkey.</sys:String>
-    <sys:String x:Key="LOCPlayState_PlaytimeSuspendRemovedResultsMessage" xml:space="preserve">Removed configuration to force to only suspend playtime when using the configured hotkey from {0} game(s).
+    <sys:String x:Key="LOCPlayState_PlaytimeSuspendAddedResultsMessage">Configured {0} game(s) to override suspend playtime only setting when using the configured hotkey.</sys:String>
+    <sys:String x:Key="LOCPlayState_PlaytimeSuspendRemovedResultsMessage" xml:space="preserve">Removed configuration to override suspend playtime only setting when using the configured hotkey from {0} game(s).
 
-Games will be suspended when using the configured hotkey.</sys:String>
+Games will respect the state of the global setting of only suspend playtime count globally and not the game processes.</sys:String>
     <sys:String x:Key="LOCPlayState_StatusSuspendedMessage">Game suspended</sys:String>
     <sys:String x:Key="LOCPlayState_StatusResumedMessage">Game resumed</sys:String>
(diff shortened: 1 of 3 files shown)


source/Generic/PlayState/PlayState.cs
```diff
@@ -120,6 +119,14 @@ public override void OnApplicationStarted(OnApplicationStartedEventArgs args)
                     .Show();
                 ProcessStarter.StartUrl(@"https://github.com/darklinkpower/PlayniteExtensionsCollection/wiki/PlayState#window-notification-style-configuration");
             }
+
+            var suspendPlaytimeOnlyFeature = PlayniteApi.Database.Features.FirstOrDefault(a => a.Name.Equals("[PlayState] Suspend Playtime only", StringComparison.OrdinalIgnoreCase));
```
darklinkpower, line 123: I'm not understanding, what is the purpose of changing the name of the feature?

source/Generic/PlayState/PlayState.cs
```diff
@@ -120,6 +119,14 @@ public override void OnApplicationStarted(OnApplicationStartedEventArgs args)
...
+            if (suspendPlaytimeOnlyFeature != null)
```
darklinkpower, line 124: Ignoring if the name should be changed or not, this check should only be done once by saving a value in settings, otherwise it will be done always on startup

source/Generic/PlayState/PlayState.cs
```diff
@@ -120,6 +119,14 @@ public override void OnApplicationStarted(OnApplicationStartedEventArgs args)
...
+            {
+                suspendPlaytimeOnlyFeature.Name = "[PlayState] Override suspend playtime only setting";
```
darklinkpower, line 126: I think the string is too long, ideally it should be short so at least most part of it can be read on the Filter Panel 
image

source/Generic/PlayState/Localization/en_US.xaml
```diff
@@ -41,7 +45,8 @@ For better compatibility with games, some adjustments to your Windows configurat
     <sys:String x:Key="LOCPlayState_SettingGlobalOnlySuspendPlaytimeLabel">Only suspend playtime count globally and not the game processes</sys:String>
     <sys:String x:Key="LOCPlayState_SettingGlobalOnlySuspendPlaytimeTooltip" xml:space="preserve">By enabling this option, all games won't be suspended when using the configured hotkey and only the time will be counted to substracted from the game playtime when it stops.
     
-This can also be applied individually per game by adding a feature named "[PlayState] Suspend playtime only" to a game</sys:String>
+The state of this setting can be overridden individually per game by adding a feature named "[PlayState] Override suspend playtime only setting" to a game.
```
darklinkpower, line 48: I think it's confusing how it would work. Instead of opposites, it should say what it does in each mode. As examples, this is done in the Special K Helper and Steam Launcher Utility extensions 
image

source/Generic/PlayState/PlayState.cs
BanCrash: Because it was intended to be only one feature instead 2, since one of them will be pointless. If you have the setting enabled, the Suspend Playtime only feature will have no effect, and if you have it disabled the new feature that I would create will not have effect.

source/Generic/PlayState/Localization/en_US.xaml
BanCrash: So let the previous feature as it is and just create a new one with the other result, right?

source/Generic/PlayState/PlayState.cs
darklinkpower: This would cause issues and create confusion when you want `X` game to suspend playtime only for any reason, like having anticheat or example. Then if you change PlayState mode, it would do the opposite and it wouldn't be clear why.

source/Generic/PlayState/Localization/en_US.xaml
darklinkpower: Yes, I think it can get very confusing otherwise. So the solution is to have two features: `[PlayState] Suspend Playtime` and `[PlayState] Suspend Processes`. This would override the current PlayState mode setting and would have the benefit of being very clear how the extension will behave with the specific game

BanCrash:
Changes done. I've removed the "Games will be suspended when using the configured hotkey." part when removing games from the Suspend playtime only feature since we can't asure that will the case.

source/Generic/PlayState/PlayState.cs
```diff
@@ -874,6 +873,24 @@ public override IEnumerable<MainMenuItem> GetMainMenuItems(GetMainMenuItemsArgs
                         var featureRemovedCount = PlayniteUtilities.RemoveFeatureFromGames(PlayniteApi, PlayniteApi.MainView.SelectedGames.Distinct(), "[PlayState] Suspend Playtime only");
                         PlayniteApi.Dialogs.ShowMessage(string.Format(ResourceProvider.GetString("LOCPlayState_PlaytimeSuspendRemovedResultsMessage"), featureRemovedCount), "PlayState");
                     }
+                },
+                new MainMenuItem
+                {
+                    Description = ResourceProvider.GetString("LOCPlayState_MenuItemAddToProcessesSuspendDescription"),
+                    MenuSection = "@PlayState",
+                    Action = a => {
+                        var featureAddedCount = PlayniteUtilities.AddFeatureToGames(PlayniteApi, PlayniteApi.MainView.SelectedGames.Distinct(), "[PlayState] Suspend Processes");
```
darklinkpower, line 882: One last thing, since both features to force a PlayState mode are contradictory, can you remove the other feature for the games while adding one of them? For example, remove `[PlayState] Suspend Playtime only"` after adding the `[PlayState] Suspend Processes` feature and also the opposite. The method should already be there in PlayniteUtilities, it's not necessary to show the removal count in the message.

darklinkpower (changes requested):

source/Generic/PlayState/PlayState.cs
darklinkpower: On second thought I'll just add it, won't take more than a minute

PR #233, opened 2022-04-12
(PlayState) Added icon and status to Manager + Fixed a potential crash with Windows notifications

BanCrash:
I have verified that:

Opening this PR because changes are starting to get big and are more changes remaining to do.

Changes on this PR:
- ~~Fixed a bug that caused Playnite to crash when trying to show a notification when the game doesn't have a banner.~~ Already added in master.
- PlayStateManager:
- - Added icon
- - Replaced "Is suspended" & "Suspend mode" with "Status"

image


darklinkpower:
I don't like the removal of these columns

darklinkpower:
I don't like or agree with the removal of these columns

image

The `IsSuspended` value gives quick info to know the state of the game. Regarding the `Status` column, that is not a replacement of the removed columns because it gives information that is different. `Suspend Mode` lets you know what will be suspended when using the extension and `IsSuspended` if it has been suspended or not.


The `Status` column you are proposing is giving information about what is the latest action done by the extension and not a status. As that is an action, I think that it should only be displayed in the notifications window as it currently does since I don't think it's something that it's useful to know afterwards.

source/Generic/PlayState/Views/PlayStateManagerView.xaml
```diff
@@ -39,26 +40,34 @@
                           ItemsSource="{Binding PlayStateDataCollection}">
                         <ListView.View>
                             <GridView>
-                                <GridViewColumn Width="250" Header="{DynamicResource LOCPlayState_ManagerGameLabel}">
+                                <GridViewColumn Width="42">
                                     <GridViewColumn.CellTemplate>
                                         <DataTemplate>
-                                            <TextBlock Text="{Binding Game.Name}" VerticalAlignment="Center"/>
+                                            <Image Height="32" Width="32" HorizontalAlignment="Center" VerticalAlignment="Center" 
+                                                   Source="{Binding Game.Icon, Converter={StaticResource IconPathConverter}}" />
```
darklinkpower, line 47: Regarding the icons loading, can you check if there are not issues deleting or replacing the icon file when it has been loaded by the window? Since it's loading the image it's probable that it will lock the image file

source/Generic/PlayState/Converters/IconPathConverter.cs
```diff
@@ -10,19 +10,22 @@
...
 namespace PlayState.Converters
 {
-    public class SuspendModeToStringConverter : MarkupExtension, IValueConverter
+    public class IconPathConverter : MarkupExtension, IValueConverter
     {
         public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
         {
-            var source = (SuspendModes)value;
-            switch (source)
+            var fullFilePath = value != null && !((string)value).IsNullOrEmpty() ? API.Instance.Database.GetFullFilePath((string)value) : null;
```
darklinkpower, line 17: I think it should be checked if the image file exists in the obtained path

source/Generic/PlayState/Views/PlayStateManagerView.xaml
BanCrash: If you use Playnite to replace or delete it, it will work as expected. If you use windows file explorer, it will happen as you say, but I think this shouldn't be the proper way to replace / delete the image file.

source/Generic/PlayState/Converters/IconPathConverter.cs
BanCrash: I thought about that but I recall that you told me that images could also be URLS, that's why I didn't add it.

BanCrash:
> I don't like or agree with the removal of these columns
> 
> image
> 
> The `IsSuspended` value gives quick info to know the state of the game. Regarding the `Status` column, that is not a replacement of the removed columns because it gives information that is different. `Suspend Mode` lets you know what will be suspended when using the extension and `IsSuspended` if it has been suspended or not.
> 
> The `Status` column you are proposing is giving information about what is the latest action done by the extension and not a status. As that is an action, I think that it should only be displayed in the notifications window as it currently does since I don't think it's something that it's useful to know afterwards.

I don't understand your comment, the new column is giving the same information as the 2 removed columns. I've just added the isSuspended part to the SuspendMode, but it's still giving the same information IMO:

image

Could you explain better what do you expect?

source/Generic/PlayState/Views/PlayStateManagerView.xaml
darklinkpower: I don't like that the file is being locked because we don't know if Playnite itself, an extension or even if the user will try to modify the image at any moment. To prevent this issue you can use the Playnite `ImageStringToImageConverter` converter, which loads the image file without locking it. Here's an example: https://github.com/darklinkpower/PlayniteExtensionsCollection/blob/a0285327cf7f312d51871d60423c13b3a4ccda6d/source/Library/JastUsaLibrary/Views/GameDownloadsView.xaml#L20-L22

source/Generic/PlayState/Converters/IconPathConverter.cs
darklinkpower: It's possible to detect if it's an URL by checking if the string starts with `http`, then only check the files that are local files

darklinkpower:
> I don't understand your comment, the new column is giving the same information as the 2 removed columns. I've just added the isSuspended part to the SuspendMode, but it's still giving the same information IMO:
> 
> image
> 
> Could you explain better what do you expect?

I see, I understand the proposition better now. How about having an option to show one or the others? Keep the current columns, add the one you are proposing and have a setting that is something like `Show game status as summary`, tooltip: `If enabled, the status of the game will be described in a single column as a summary`. Or something similar

source/Generic/PlayState/Converters/IconPathConverter.cs
darklinkpower: If you want to skip icons that are url then that's also fine for me

PR #301, opened 2022-08-16
GOG size calc, Steam game size calc improvements

Jeshibu:
I have verified that:


source/Generic/GamesSizeCalculator/GamesSizeCalculator.cs
```diff
@@ -279,16 +295,38 @@ public override void OnLibraryUpdated(OnLibraryUpdatedEventArgs args)
...
+            var onlineSizeCalculators = new List<IOnlineSizeCalculator>();
+            if (settings.Settings.GetUninstalledGameSizeFromSteam)
+            {
+                onlineSizeCalculators.Add(new SteamSizeCalculator(steamClient, GetDefaultSteamAppUtility()
+                    , settings.Settings.IncludeDlcInSteamCalculation
+                    , settings.Settings.IncludeOptionalInSteamCalculation
+                    , settings.Settings.GetSizeFromSteamNonSteamGames
+                    , settings.Settings.DepotRegionWords.ToArray()
+                    , settings.Settings.DepotRegionWordsBlacklist.ToArray()));
+            }
```
darklinkpower, line 309: Is it possible that instead of passing all those settings values one by one, pass the whole settings object instead and handle it in the constructor to keep it cleaner perhaps?

source/Generic/GamesSizeCalculator/GamesSizeCalculator.cs
```diff
@@ -279,16 +295,38 @@ public override void OnLibraryUpdated(OnLibraryUpdatedEventArgs args)
...
+        }
+
         private void ProcessGamesOnLibUpdate(GlobalProgressActionArgs a)
         {
-            var games = PlayniteApi.Database.Games.Where(x => x.IsInstalled);
-            a.ProgressMaxValue = games.Count();
+            ICollection<Game> games = PlayniteApi.Database.Games;
+            if (!settings.Settings.GetUninstalledGameSizeFromSteam)
+                games = PlayniteApi.Database.Games.Where(x => x.IsInstalled).ToList();
+
```
darklinkpower, line 322: Use curly braces here please. Perhaps it could be done in one line like this

`ICollection<Game> games = settings.Settings.GetUninstalledGameSizeFromSteam ? PlayniteApi.Database.Games : PlayniteApi.Database.Games.Where(x => x.IsInstalled).ToList()`

source/Generic/GamesSizeCalculator/GamesSizeCalculator.csproj
```diff
@@ -34,6 +34,9 @@
     <Reference Include="AngleSharp, Version=0.9.9.0, Culture=neutral, PublicKeyToken=e83494dcdc6d31ea, processorArchitecture=MSIL">
       <HintPath>..\..\packages\AngleSharp.0.9.9\lib\net45\AngleSharp.dll</HintPath>
     </Reference>
+    <Reference Include="Newtonsoft.Json, Version=10.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Newtonsoft.Json.10.0.3\lib\net45\Newtonsoft.Json.dll</HintPath>
+    </Reference>
```
darklinkpower, line 39: As mentioned before, if possible use the `Serialization` class to eliminate the dependency

source/Generic/GamesSizeCalculator/GamesSizeCalculatorSettingsView.xaml
```diff
@@ -6,37 +6,84 @@
...
+                                  ToolTip="{DynamicResource LOCGame_Sizes_Calculator_SettingsIncludeOptionalInSteamCalculationTooltip}"
+                                  IsChecked="{Binding Settings.IncludeOptionalInSteamCalculation}"/>
+                    </StackPanel>
+                    <CheckBox Content="For uninstalled PC games, obtain an estimated install size from GOG.com"
+                              Margin="0,10,0,0"
+                              IsChecked="{Binding Settings.GetUninstalledGameSizeFromGog}"/>
+                    <StackPanel Margin="40,0,0,0"
+                                IsEnabled="{Binding Settings.GetUninstalledGameSizeFromGog}">
+                        <CheckBox Content="Apply this data also for games that are not from the GOG library"
+                                  ToolTip="Games must have a link to the GOG store page"
```
darklinkpower, line 49: Add localization strings for new strings please

source/Generic/GamesSizeCalculator/GamesSizeCalculatorSettings.cs
```diff
@@ -25,6 +26,42 @@ public class GamesSizeCalculatorSettings : ObservableObject
...
+
+        private ObservableCollection<string> depotRegionWords = new ObservableCollection<string> { "eu", "europe", "row", "en", "english", "ww" };
+        public ObservableCollection<string> DepotRegionWords { get => depotRegionWords; set => SetValue(ref depotRegionWords, value); }
+
+        private ObservableCollection<string> depotRegionWordsBlacklist = new ObservableCollection<string>
+        {
+            "asia",
+            "aus",
+            "australia",
+            "nz",
```
darklinkpower, line 44: Is this collection modifiable by the user or will it never change? If it never changes then it should be moved to the settings Viewmodel object instead as there wouldn't be a purpose in these being here. If they are modifiable then ignore this

source/Generic/GamesSizeCalculator/GamesSizeCalculatorSettingsView.xaml
```diff
@@ -6,37 +6,84 @@
...
-                      IsChecked="{Binding Settings.GetSizeFromSteamNonSteamGames}"/>
-            <CheckBox Content="{DynamicResource LOCGame_Sizes_Calculator_SettingsIncludeDlcInSteamCalculationLabel}"
-                      Margin="0,10,0,0"
-                      IsChecked="{Binding Settings.IncludeDlcInSteamCalculation}"/>
-            <CheckBox Content="{DynamicResource LOCGame_Sizes_Calculator_SettingsIncludeOptionalInSteamCalculationLabel}"
-                      Margin="0,10,0,0"
-                      ToolTip="{DynamicResource LOCGame_Sizes_Calculator_SettingsIncludeOptionalInSteamCalculationTooltip}"
-                      IsChecked="{Binding Settings.IncludeOptionalInSteamCalculation}"/>
-        </StackPanel>
+    <TabControl ScrollViewer.VerticalScrollBarVisibility="Disabled" HorizontalContentAlignment="Stretch" VerticalContentAlignment="Stretch">
```
darklinkpower, line 9: Use a Panel as the main control here please. Either `Grid` or `StackPanel` with a margin of `20` for consistency and place the `TabControl as a child`. Also what is the purpose of `ScrollViewer.VerticalScrollBarVisibility` and the ContenAlignment properties?

source/Generic/GamesSizeCalculator/GamesSizeCalculatorSettingsView.xaml
```diff
@@ -6,37 +6,84 @@
...
 
-    </StackPanel>
+                <Grid VerticalAlignment="Stretch" Margin="0,10,0,0">
+                    <Grid.ColumnDefinitions>
+                        <ColumnDefinition x:Name="Left" Width="*"/>
+                        <ColumnDefinition x:Name="Middle" Width="50"/>
+                        <ColumnDefinition x:Name="Right" Width="*"/>
+                    </Grid.ColumnDefinitions>
+                    <Grid.RowDefinitions>
+                        <RowDefinition x:Name="Header" MaxHeight="20"/>
```
darklinkpower, line 70: The Row shouldn't have a `MaxHeight` because it will cause issues if the text is taller, which can happen in bigger fonts. Imo it should be set to `Auto`

source/Generic/GamesSizeCalculator/GamesSizeCalculatorSettingsView.xaml
```diff
@@ -6,37 +6,84 @@
...
+                        <RowDefinition x:Name="Body"/>
+                    </Grid.RowDefinitions>
+                    <TextBlock HorizontalAlignment="Center" Grid.Column="0" Grid.Row="0">Region-words</TextBlock>
+                    <TextBlock HorizontalAlignment="Center" Grid.Column="2" Grid.Row="0">Excluded region-words</TextBlock>
+                    <TextBox DockPanel.Dock="Top" Name="Whitelist"
```
darklinkpower, line 75: Why does it have a `DockPanel.Dock` property if the parent control is a `Grid` ?

source/Generic/GamesSizeCalculator/Steam/SteamSizeCalculator.cs
```diff
@@ -9,84 +10,160 @@
 
 namespace GamesSizeCalculator.SteamSizeCalculation
 {
-    public class SteamSizeCalculator
+    public class SteamSizeCalculator : IOnlineSizeCalculator
     {
         private ILogger logger = LogManager.GetLogger();
```
darklinkpower, line 15: This one can be set as static

source/Generic/GamesSizeCalculator/Steam/SteamSizeCalculator.cs
```diff
@@ -9,84 +10,160 @@
...
         public ISteamApiClient SteamApiClient { get; }
+        public ISteamAppIdUtility SteamAppIdUtility { get; }
+        public bool IncludeDLC { get; set; }
+        public bool IncludeOptional { get; set; }
+        public bool HandleNonSteamGames { get; set; }
+        public string[] RegionalWords { get; set; }
+        public string[] RegionalWordsBlacklist { get; set; }
+        public string ServiceName { get; } = "Steam";
```
darklinkpower, line 23: Should be a constant. Also properties that are not change later should be set to `readonly`

source/Generic/GamesSizeCalculator/Steam/SteamSizeCalculator.cs
```diff
@@ -9,84 +10,160 @@
...
-        public SteamSizeCalculator(ISteamApiClient steamApiClient)
+        public SteamSizeCalculator(ISteamApiClient steamApiClient, ISteamAppIdUtility steamAppIdUtility, bool includeDLC, bool includeOptional, bool handleNonSteamGames, string[] regionalWords, string[] regionalWordsBlacklist)
         {
             SteamApiClient = steamApiClient;
+            SteamAppIdUtility = steamAppIdUtility;
+            IncludeDLC = includeDLC;
+            IncludeOptional = includeOptional;
+            HandleNonSteamGames = handleNonSteamGames;
+            RegionalWords = regionalWords;
+            RegionalWordsBlacklist = regionalWordsBlacklist;
```
darklinkpower, line 33: I'm wondering if instead of setting individual fields from settings, it would be possible to set a settings field with the plugin settings and access the properties directly

source/Generic/GamesSizeCalculator/Steam/SteamSizeCalculator.cs
```diff
@@ -9,84 +10,160 @@
...
+            return size;
+        }
+
+        private class DepotGroupingInfo
+        {
+            public string BaseName { get; set; }
+            public string RegionWord { get; set; }
+            public int Rank { get; set; }
+            public DepotInfo Depot { get; set; }
         }
```
darklinkpower, line 89: Models should be in a separate class dedicated to model(s)

source/Generic/GamesSizeCalculator/Steam/SteamSizeCalculator.cs
```diff
@@ -9,84 +10,160 @@
...
-                        logger.Debug($"Removed depot {depot.Name}");
-                        break;
-                    default:
-                        break;
+                    var depotData = orderedDepots[i];
+                    string depotLogString = $"{depotData.Depot.Id} - {depotData.Depot.Name}, optional: {depotData.Depot.Optional}, dlc: {depotData.Depot.IsDLC}, size: {depotData.Depot.FileSize}, {depotData.BaseName}|{depotData.RegionWord}|{depotData.Rank}";
+
+                    if (i == 0)
+                    {
+                        logString += "Keeping depot: " + depotLogString;
```
darklinkpower, line 121: Imo a StringBuilder should be used instead of recreating the string with each addition

source/Generic/SteamGameTransferUtility/ViewModels/TransferWindowViewModel.cs
```diff
@@ -356,14 +356,14 @@ public string CalculateSize(string directory)
...
+        private static string FormatBytes(ulong bytes)
         {
             string[] Suffix = { "B", "KB", "MB", "GB", "TB" };
             int i;
             double dblSByte = bytes;
-            for (i = 0; i < Suffix.Length && bytes >= 1024; i++, bytes /= 1024)
+            for (i = 0; i < Suffix.Length && bytes >= 1024UL; i++, bytes /= 1024UL)
             {
-                dblSByte = bytes / 1024.0;
+                dblSByte = bytes / 1024D;
```
darklinkpower, line 366: Should not be changed :(

darklinkpower (changes requested):

source/Generic/GamesSizeCalculator/GOG/GogApiClient.cs
```diff
@@ -0,0 +1,101 @@
...
+                if (line.TrimStart().StartsWith("window.productcardData"))
+                {
+                    dataStarted = true;
+                    stringData = trimmed.Substring(25).TrimEnd(';');
+                    continue;
+                }
+
+                if (line.TrimStart().StartsWith("window.activeFeatures"))
+                {
+                    var desData = Newtonsoft.Json.JsonConvert.DeserializeObject<StorePageResult>(stringData.TrimEnd(';'));
```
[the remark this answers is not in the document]
Jeshibu: Yes, Playnite's deserialization throws exceptions when using them in a unit test.

source/Generic/GamesSizeCalculator/GamesSizeCalculatorSettingsView.xaml
Jeshibu: Why an extra wrapping element rather than a Margin="20" for the TabControl?

The properties are for the second tab, I'll move them there.

source/Generic/GamesSizeCalculator/GOG/GogApiClient.cs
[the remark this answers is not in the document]
Jeshibu: Playnite's deserialization requires Playnite to be initialized and is unusable for unit tests.

source/Generic/GamesSizeCalculator/GOG/GogApiClient.cs
```diff
@@ -0,0 +1,101 @@
...
+                }
+            }
+
+            logger.Warn("Failed to get store data from page, no data found. " + gameUrl);
+            return null;
+        }
+
+        public ProductApiDetail GetGameDetails(string id)
+        {
+            var baseUrl = @"https://api.gog.com/products/{0}?expand=description";
```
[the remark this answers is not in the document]
Jeshibu: converted to inline constant since it's not used elsewhere

source/Generic/GamesSizeCalculator/Steam/SteamSizeCalculator.cs
Jeshibu: Unclear what you mean here

source/Generic/GamesSizeCalculator/Steam/SteamSizeCalculator.cs
Jeshibu: It's an interface member, so it can't be static, and properties can't be readonly in C# 7.3

source/Generic/GamesSizeCalculator/GamesSizeCalculatorSettingsView.xaml
Jeshibu: Because its grandparent is a dockpanel. Does this do nothing or something?

source/Generic/GamesSizeCalculator/Steam/SteamSizeCalculator.cs
darklinkpower: That's true, I didn't notice

source/Generic/GamesSizeCalculator/Steam/SteamSizeCalculator.cs
darklinkpower: private static ILogger logger = LogManager.GetLogger();

But it can be kept this way too

source/Generic/GamesSizeCalculator/GamesSizeCalculator.cs
```diff
@@ -279,16 +302,35 @@ public override void OnLibraryUpdated(OnLibraryUpdatedEventArgs args)
...
+                onlineSizeCalculators.Add(new GOG.GogSizeCalculator(new GOG.HttpDownloaderWrapper(), settings.Settings.GetSizeFromGogNonGogGames));
+            }
+            return onlineSizeCalculators;
+        }
+
         private void ProcessGamesOnLibUpdate(GlobalProgressActionArgs a)
         {
-            var games = PlayniteApi.Database.Games.Where(x => x.IsInstalled);
-            a.ProgressMaxValue = games.Count();
+            ICollection<Game> games = PlayniteApi.Database.Games;
```
darklinkpower, line 321: What's the purpose of using `ICollection` here?

source/Generic/GamesSizeCalculator/GamesSizeCalculatorSettingsView.xaml
```diff
@@ -6,37 +6,79 @@
...
+                        <RowDefinition x:Name="Body"/>
+                    </Grid.RowDefinitions>
+                    <TextBlock HorizontalAlignment="Center" Grid.Column="0" Grid.Row="0" Text="{DynamicResource LOCGame_Sizes_Calculator_SettingsRegionWordsHeaderLabel}"/>
+                    <TextBlock HorizontalAlignment="Center" Grid.Column="2" Grid.Row="0" Text="{DynamicResource LOCGame_Sizes_Calculator_SettingsExcludedRegionWordsHeaderLabel}"/>
+                    <TextBox DockPanel.Dock="Top" Name="Whitelist" AcceptsReturn="True"
+                                MinHeight="220"
+                                VerticalAlignment="Stretch"
+                                Text="{Binding RegionWordsString}"
+                                Grid.Column="0" Grid.Row="1"/>
+                    <TextBox DockPanel.Dock="Top" Name="Blacklist" AcceptsReturn="True"
```
darklinkpower, line 75: Remove the `DockPanel.Dock` please as they don't do anything

source/Generic/GamesSizeCalculator/Localization/en_US.xaml
```diff
@@ -1,15 +1,27 @@
...
+    <sys:String x:Key="LOCGame_Sizes_Calculator_SettingsApplyGOGSizesToOtherLibrariesLabel">Apply this data also for games that are not from the GOG library</sys:String>
+    <sys:String x:Key="LOCGame_Sizes_Calculator_SettingsApplyGOGSizesToOtherLibrariesTooltip">Games must have a link to the GOG store page</sys:String>
+
+    <sys:String x:Key="LOCGame_Sizes_Calculator_SettingsRegionWordsExplanationText">Region-words are things that denote regions for otherwise identically named Steam depots. For example, "Content EU" and "Content Asia" would have the last word stripped and only one would remain if EU and Asia were both in the Region-words textbox here. If you're sure you'll never need a region, enter it into the Excluded region-words textbox and depots ending in that word wil always be exlcuded from size calculation. Put each word on its own line.</sys:String>
+    <sys:String x:Key="LOCGame_Sizes_Calculator_SettingsRegionWordsHeaderLabel">Region-words</sys:String>
+    <sys:String x:Key="LOCGame_Sizes_Calculator_SettingsExcludedRegionWordsHeaderLabel">Excluded region-words</sys:String>
```
darklinkpower, line 23: In my opinion `Region-words` used in the different localization strings is not clear enough and can be confusing, would it be possible to use something else instead? Something like `Excluded regions by word`. For the tab header, maybe `Regions data import settings` or something like that that is in my opinion more clear and eloquent.

Jeshibu: I prefer to leave that to your judgement after merging, though I think "excluded regions by word" makes it sound more magical than it really is

darklinkpower:
Closed because I don't like you

PR #309, opened 2022-09-04
[PlayState] Added reminder if currentGame is paused

### LINKED ISSUE #257: [PlayState] Add a setting to remember each x minutes that you have paused your current game
### Extension name

PlayState

### Feature Description

Hi, I find very useful having a setting that if it's enabled when you have your current game paused for x minutes, it will send a notification to remember you that it's paused. So in case it's no longer paused (in case you are only pausing playtime) but you forgot to resume you can do it after seeing the notification.

I have this feature in my previous game launcher and it's funny how many times I resumed a game that I was already playing thanks to the notification because I forgot to do so. So for me it's very useful.

### Screenshots

_No response_

BanCrash:
I have verified that:

-Closes #257 
Opening as draft since I didn't include the settings yet since I want to know first if you like this approach.
The idea is the settings will have an option to enable or disable this, and a slider to choose how often (in minutes) will reminder that the currentGame is paused.

### DIFF
--- a/source/Generic/PlayState/Models/PlayStateData.cs
+++ b/source/Generic/PlayState/Models/PlayStateData.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using System.Windows.Threading;
 
 namespace PlayState.Models
 {
@@ -20,6 +21,8 @@ namespace PlayState.Models
         public DateTime StartDate { get; } = DateTime.Now;
         private Stopwatch stopwatch = new Stopwatch();
         public Stopwatch Stopwatch { get => stopwatch; set => SetValue(ref stopwatch, value); }
+        private DispatcherTimer reminderTimer = new DispatcherTimer();
+        public DispatcherTimer ReminderTimer { get => reminderTimer; set => SetValue(ref reminderTimer, value); }
         public List<ProcessItem> GameProcesses { get; set; }
         public bool HasProcesses => GameProcesses?.HasItems() == true;
 
--- a/source/Generic/PlayState/Models/PlayStateData.cs
+++ b/source/Generic/PlayState/Models/PlayStateData.cs
@@ -89,5 +89,19 @@ namespace PlayState.Models
                 GameProcesses = null;
             }
         }
+
+        public void SetReminderTimer()
+        {
+            if (ReminderTimer.IsEnabled)
+            {
+                ReminderTimer.Stop();
+            }
+            else
+            {
+                ReminderTimer = new DispatcherTimer();
+                ReminderTimer.Interval = TimeSpan.FromMinutes(1);
+                ReminderTimer.Start();
+            }
+        }
     }
 }
--- a/source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
+++ b/source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
@@ -415,7 +415,6 @@ namespace PlayState.ViewModels
                 }
 
                 var notificationType = NotificationTypes.None;
-                var reminder = true;
                 if (processesSuspended || gameData.SuspendMode == SuspendModes.Playtime)
                 {
                     if (gameData.IsSuspended)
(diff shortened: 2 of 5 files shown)


source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
```diff
@@ -415,12 +424,23 @@ public bool SwitchGameState(PlayStateData gameData)
...
                     {
                         gameData.IsSuspended = true;
                         notificationType = processesSuspended ? NotificationTypes.Suspended : NotificationTypes.PlaytimeSuspended;
                         gameData.Stopwatch.Start();
+                        if (reminder)
+                        {
+                            gameData.ReminderTimer = new DispatcherTimer();
+                            gameData.ReminderTimer.Interval = TimeSpan.FromMinutes(1);
+                            gameData.ReminderTimer.Tick += (sender, e) => ShowNotificationReminderIfCurrentGameIsSuspended(gameData);       
+                            gameData.ReminderTimer.Start();
```
darklinkpower, line 442: I don't like that the timer is being created here, outside of the object itself. I think there are two options:

- Have a single timer in the `PlayStateManager` that checks all Data objects and sends notifications if needed. Easiest to implement.
- Have a timer created inside the `PlayStateData`. It would also need an Event that sends a signal that a notification should be sent. The PlayStateManager subscribes to this event. Remember that events should be unsubscribed when disposing the object

BanCrash: @darklinkpower I moved the reminder to PlayStateData, though it still add the method call in the viewmodel in order to make it simpler. Is this approach fine for you?

source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
```diff
@@ -415,12 +423,21 @@ public bool SwitchGameState(PlayStateData gameData)
...
+                        {
+                            gameData.SetReminderTimer();
+                        }
                     }
                     else
                     {
                         gameData.IsSuspended = true;
                         notificationType = processesSuspended ? NotificationTypes.Suspended : NotificationTypes.PlaytimeSuspended;
                         gameData.Stopwatch.Start();
+                        if (settings.Settings.NotificationShowSuspendedReminder)
```
darklinkpower, line 436: I think the functionality should be able to be enable for both suspended and not suspended games, in different options for more flexibility and not only for suspended games

source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
```diff
@@ -415,12 +423,21 @@ public bool SwitchGameState(PlayStateData gameData)
...
+                        {
+                            gameData.SetReminderTimer();
+                            gameData.ReminderTimer.Tick += (sender, e) => ShowNotificationReminderIfCurrentGameIsSuspended(gameData);
```
darklinkpower, line 439: This should be unsubscribed when disposing of the `PlayStateData` object

darklinkpower (changes requested):

source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
darklinkpower: I'm also not sure that the subscribing and everything needed should be in this method, why is it done here?

source/Generic/PlayState/Models/PlayStateData.cs
```diff
@@ -86,5 +89,19 @@ internal void RemoveProcesses()
...
+        public void SetReminderTimer()
+        {
+            if (ReminderTimer.IsEnabled)
+            {
+                ReminderTimer.Stop();
+            }
+            else
+            {
+                ReminderTimer = new DispatcherTimer();
+                ReminderTimer.Interval = TimeSpan.FromMinutes(1);
```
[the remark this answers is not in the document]
BanCrash: Yes, I just added the enabler setting because I was moving to the other PR, but my original idea was to add both at the same time.

source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
BanCrash: You mean a reminder that reminds you that you been playing a game for x minutes, right?
If so, should the reminder time be the same for both scenarios, or be configurable for each case? (example, 2 minutes reminder for suspended game, 30 minutes for played game).

BanCrash: > I'm also not sure that the subscribing and everything needed should be in this method, why is it done here?

The idea is that when you pause the game, after for example 5 minutes you'll get the notification. If you resume the game, after the second pause the reminder should start from 0. If there is another suspension trigger I could move the method there  but since the Soptwatch was started there too for me was logical to add it there too.

BanCrash:
Superseded by #357

PR #322, opened 2022-10-09
[PlayState] Start monitoring faster games when they will be only suspended by playtime

BanCrash:
I have verified that:

Let's see what do you think about this. With this PR, when a game is set to only suspend playtime, it will start monitoring it faster since it will start the detection before going to the big loop. 

Explanation:
Without this PR, Valorant takes 3 minutes to get detected because PlayState for some reason it's not detecting their processes. After this PR, on 20 seconds it's started to monitoring, without processes detected. If eventually some process is found in the loop, it will get added to the game anyway. This only apply for the case of suspend playtime games.

Maybe could be added a setting to enable or disable this.

### DIFF
--- a/source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
+++ b/source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
@@ -195,10 +195,13 @@ namespace PlayState.ViewModels
             var data = GetDataOfGame(game);
             if (data != null)
             {
+                // Add processes to the game if the data for this game already exists but doesn't have any processes
+                // This is for games that are added without game processes before going to the WMI loop, but after the loop some processes are detected
                 if (!data.HasProcesses && gameProcesses?.HasItems() == true)
                 {
-                    data.GameProcesses = gameProcesses;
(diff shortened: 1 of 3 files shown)


source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
```diff
@@ -189,16 +189,28 @@ internal void AddPlayStateData(Game game, List<ProcessItem> gameProcesses, bool
...
+
+            var data = GetDataOfGame(game);
+            if (data != null)
+            {
+                if (!data.HasProcesses && gameProcesses?.HasItems() == true)
+                {
+                    data.GameProcesses = gameProcesses;
+                    logger.Debug($"Processes for game {game.Name} with id {game.Id} were added. Executables: {procsExecutablePaths}");
+                }
+                else
```
darklinkpower, line 203: Can you add a comment of why processes are being added here please?

darklinkpower (changes requested):

BanCrash:
Also added the SetProcesses method to PlayStateData.

PR #329, opened 2022-10-22
[PlayState] Added suspension / resume override for manual suspension status switch

### LINKED ISSUE #306: [PlayState] Have a setting to choose the behaviour when starting a second game
### Extension name

PlayState

### Feature Description

Hi, when launching a second (third, etc) game, I think a setting to choose the behaviour to the previous launched game will be useful. The idea is to have a setting where you can choose:
- Do nothing -> Keep the actual state of the previous launched game.
- Pause game -> Pause the game regardless the state of the game (if it's already paused do nothing).
- Resume game -> Resume the game regardless the state of the game (if it's already resumed do nothing)

I'm asking because I want that when I launch a second (or more) game, the previous game get automatically paused, but I think that other people would want other behaviour so the best option is to let the user choose what they want to do with their games when a new one is launched.

### Screenshots

_No response_

BanCrash:
I have verified that:

Current behaviour before this PR:
- When you have enabled the option of automatically suspend / resume a game if is / is not in the foreground, if you manually suspend / resume the game, on the next second it will automatically get reverted.

After this PR:
- If you manually suspend / resume a game, this setting will be overrided until you minimize and restore the game again, so if you pause a game that is on the foreground, it won't get resume unless you manually do so again or after you minimize it and set again to the foreground.

I think this is the expected behaviour, because with the current behaviour manually suspending or resuming a game doesn't have any real impact if you have that setting enabled.

IMO this partially closes #306. Only thing remaining for fully closing that issue would be add support for manually choose the process to watch, so games like Valorant could be tracked with the automatic option, since currently PlayState is not detecting any game processes.

### DIFF
--- a/source/Generic/PlayState/Enums/GameState.cs
+++ b/source/Generic/PlayState/Enums/GameState.cs
@@ -1,8 +1,0 @@
-﻿namespace PlayState.Enums
-{
-    public enum PlayStateGameState
-    {
-        Resumed = 1,
-        Paused = 2
-    }
-}
--- a/source/Generic/PlayState/Enums/GameState.cs
+++ b/source/Generic/PlayState/Enums/GameState.cs
@@ -1,0 +1,8 @@
+﻿namespace PlayState.Enums
+{
+    public enum PlayStateGameState
+    {
+        Resumed = 1,
+        Paused = 2
+    }
+}
--- a/source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
+++ b/source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
@@ -122,8 +122,8 @@ namespace PlayState.ViewModels
                 }
 
                 if (playstateData.IsGameStatusOverrided && 
-                    (isForeground && playstateData.GameStatusOverride == PlayStateGameState.Resumed ||
-                    !isForeground && playstateData.GameStatusOverride == PlayStateGameState.Paused))
+                    (isForeground && playstateData.GameStatusOverride == PlayStateAutomaticStateSwitchStatus.Enabled ||
+                    !isForeground && playstateData.GameStatusOverride == PlayStateAutomaticStateSwitchStatus.Disabled))
                 {
                     playstateData.IsGameStatusOverrided = false;
                 }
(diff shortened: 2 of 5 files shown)


source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
```diff
@@ -415,13 +430,23 @@ public bool SwitchGameState(PlayStateData gameData)
...
                     else
                     {
                         gameData.IsSuspended = true;
                         notificationType = processesSuspended ? NotificationTypes.Suspended : NotificationTypes.PlaytimeSuspended;
                         gameData.Stopwatch.Start();
                         logger.Debug($"Game {gameData.Game.Name} suspended in mode {gameData.SuspendMode}");
+                        if (isGameStatusOverrided)
+                        {
+                            gameData.IsGameStatusOverrided = true;
+                            gameData.GameStatusOverride = PlayStateDataStatus.Paused;
```
darklinkpower, line 448: I think it should use enums specific for this use case instead of the ones intended for something else

darklinkpower (changes requested):

source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
BanCrash: Sure, I did that way to avoid adding more code but it's okay for me.

BanCrash: Done

source/Generic/PlayState/Enums/GameState.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace PlayState.Enums
+{
+    public enum PlayStateGameState
```
darklinkpower, line 3: I think the name of the enum and its values are not very clear in what they do or what they purpose is. Can you change it to something else please? Maybe something like `AutomaticStateSwitchStatus` with `Enabled` and `Disabled` values but if you have a better idea let me know

darklinkpower (changes requested):

source/Generic/PlayState/Enums/GameState.cs
BanCrash: Sure, fine for me.

BanCrash: Done.

PR #350, opened 2022-12-14
Support for ExpandableVariables

### LINKED ISSUE #344: [Extra Metadata Loader] Relative Directories
### Extension name

Extra Metadata Loader

### Feature Description

Perhaps it is already an feature and I'm missing it, but is it possible to use relative directories for the FFMPEG, FFPROBE, and YT-DLP executables?

I use Playnite as a portable application, but if I transfer my portable drive to a different computer it is very possible the drive letter will be different.  I've currently placed the executables in "F:\Playnite\Extensions\ExtraMetadataLoader_705fdbca-e1fc-4004-b839-1d040b8b4429\Extras" and the extension is configured to that directory.  However plugging the device in my son's computer changes the drive letter to E:\ and the extension ceases to work.

Is there a variable, or can I even use a variable, to point to the Playnite folder.  I know that {PlayniteDir} is an existing variable, but you can't directly add that to the text fields.

Thanks in advance for your response.

### Screenshots

_No response_

maforget:
This fixes #344, there are no support to change the path in the GUI. I just change the path in the `config.json` directly. Also since the API uses the Game for some Variables and some of the methods don't supply it, I just `new Game()` instead in that case. So it will work with `{PlayniteDir}`, but might not always work with other Variables that require the Game object.

Did this for personal use, feel free to use it or modify it if required.

I have verified that:

### DIFF
--- a/source/Generic/ExtraMetadataLoader/ExtraMetadataLoader.cs
+++ b/source/Generic/ExtraMetadataLoader/ExtraMetadataLoader.cs
@@ -896,33 +896,37 @@ namespace ExtraMetadataLoader
 
         private bool ValidateExecutablesSettings(bool validateFfmpeg, bool validateYtdl)
         {
+            var youtubeDlPath = extraMetadataHelper.ExpandVariables(settings.Settings.YoutubeDlPath);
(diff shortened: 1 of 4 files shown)


darklinkpower:
Please also make the paths text field editable from the extension settings page

maforget:
Done

PR #357, opened 2023-01-08
[PlayState] Added playing & suspension reminders

### LINKED ISSUE #257: [PlayState] Add a setting to remember each x minutes that you have paused your current game

BanCrash:
I have verified that:

-Supersede #309
-Closes #257

- Added a notification after x minutes of suspension of the game.
- Added a notification afer x minutes of playing the game.
- Added settings for enabling and disabling this and for the duration of each setting.

Opening as a draft since I would like to give it more testing, but the code should be complete.

### DIFF
--- a/source/Generic/PlayState/Models/PlayStateData.cs
+++ b/source/Generic/PlayState/Models/PlayStateData.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using System.Windows.Threading;
 
 namespace PlayState.Models
 {
@@ -20,6 +21,10 @@ namespace PlayState.Models
         public DateTime StartDate { get; } = DateTime.Now;
         private Stopwatch stopwatch = new Stopwatch();
         public Stopwatch Stopwatch { get => stopwatch; set => SetValue(ref stopwatch, value); }
+        private DispatcherTimer suspensionReminderTimer = new DispatcherTimer();
+        public DispatcherTimer SuspensionReminderTimer { get => suspensionReminderTimer; set => SetValue(ref suspensionReminderTimer, value); }
+        private DispatcherTimer playingReminderTimer = new DispatcherTimer();
+        public DispatcherTimer PlayingReminderTimer { get => playingReminderTimer; set => SetValue(ref playingReminderTimer, value); }
         public List<ProcessItem> GameProcesses { get; set; }
         public bool HasProcesses => GameProcesses?.HasItems() == true;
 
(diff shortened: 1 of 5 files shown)


source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
```diff
@@ -444,6 +460,9 @@ public bool SwitchGameState(PlayStateData gameData, bool isGameStatusOverrided =
                         gameData.IsSuspended = false;
                         notificationType = processesSuspended ? NotificationTypes.Resumed : NotificationTypes.PlaytimeResumed;
                         gameData.Stopwatch.Stop();
+                        gameData.StartPlayingReminderTimer();
+                        gameData.PlayingReminderTimer.Tick += (sender, e) => ShowPlayingReminderNotificationIfGameIsCurrentGame(gameData);
```
darklinkpower, line 464: This subscription will cause issues because there won't be a way to unsubscribe later when disposing the object. Also this being here means that subscriptions will be made each time the method is called, which is not good

source/Generic/PlayState/ViewModels/PlayStateManagerViewModel.cs
```diff
@@ -456,6 +475,9 @@ public bool SwitchGameState(PlayStateData gameData, bool isGameStatusOverrided =
                         gameData.IsSuspended = true;
                         notificationType = processesSuspended ? NotificationTypes.Suspended : NotificationTypes.PlaytimeSuspended;
                         gameData.Stopwatch.Start();
+                        gameData.StopPlayingReminderTimer();
+                        gameData.StartSuspensionReminderTimer();
+                        gameData.SuspensionReminderTimer.Tick += (sender, e) => ShowSuspensionReminderNotificationIfGameIsCurrentGame(gameData);
```
darklinkpower, line 480: Same here

source/Generic/PlayState/Models/PlayStateData.cs
```diff
@@ -58,6 +63,14 @@ private void Settings_PropertyChanged(object sender, System.ComponentModel.Prope
...
                 SetSuspendMode();
             }
+            if (e.PropertyName == nameof(settingsModel.Settings.NotificationPlayingReminderMinutes))
+            {
+                PlayingReminderTimer.Interval = TimeSpan.FromMinutes(settingsModel.Settings.NotificationPlayingReminderMinutes);
+            }
+            if (e.PropertyName == nameof(settingsModel.Settings.NotificationSuspensionReminderMinutes))
+            {
+                SuspensionReminderTimer.Interval = TimeSpan.FromMinutes(settingsModel.Settings.NotificationSuspensionReminderMinutes);
+            }
```
darklinkpower, line 73: I don't like this design, I think the object itself shouldn't take care of what happens outside of it. The `PlayStateData` object should only be in charge of holding data and not have functionality. In my opinion the `PlayState Manager` should be in charge of handling the notification events. If possible please design this another way

BanCrash:
Since I don't have time currently with studying and working, and anyway when I have time to do this will be done on a new PR, I'll close this and leave on the drawer until I have the free time to make the suggested changes.

PR #359, opened 2023-01-21
Added MicroTrailer support

cnapolit:
Will only use micro trailer if settings allow it, the file exists, & the user has not manually specified a file.
Also, some minor refactors in the relevant code because why not.

Tested each of the following options for a total of 8 test cases:
- Playnite Mode state (Desktop or Fullscreen)
- UseMicroTrailerFullscreenMode
- UseMicroTrailerDesktopMode

Also tested:
- Intro video & Micro trailer files do not exist
- Intro video & Micro trailer files exist

### DIFF
--- a/source/Generic/SplashScreen/SplashScreen.cs
+++ b/source/Generic/SplashScreen/SplashScreen.cs
@@ -331,48 +331,59 @@ namespace SplashScreen
 
         private string GetSplashVideoPath(Game game)
         {
-            var videoPathTemplate = Path.Combine(PlayniteApi.Paths.ConfigurationPath, "ExtraMetadata", "{0}", "{1}", "VideoIntro.mp4");
+            const string videoIntroName = "VideoIntro.mp4";
+            const string microVideoName = "VideoMicrotrailer.mp4";
 
-            var splashVideo = string.Format(videoPathTemplate, "games", game.Id.ToString());
+            var baseVideoPathTemplate = Path.Combine(PlayniteApi.Paths.ConfigurationPath, "ExtraMetadata", "{0}", "{1}");
+            
+            var baseSplashVideo = string.Format(baseVideoPathTemplate, "games", game.Id.ToString());
+            var splashVideo = Path.Combine(baseSplashVideo, videoIntroName);
             if (FileSystem.FileExists(splashVideo))
             {
                 logger.Info(string.Format("Specific game video found in {0}", splashVideo));
-                return splashVideo;
+                return LogAcquiredVideo("Game-specific", splashVideo);
             }
+
+            var isDesktop = PlayniteApi.ApplicationInfo.Mode is ApplicationMode.Desktop;
+            if (ShouldUseMicroTrailer(isDesktop))
+            {
+                splashVideo = Path.Combine(baseSplashVideo, microVideoName);
+                if (FileSystem.FileExists(splashVideo))
+                {
+                    return LogAcquiredVideo("Micro trailer", splashVideo);
+                }
+            }
+
+            var videoPathTemplate = Path.Combine(baseVideoPathTemplate, videoIntroName);
 
             if (game.Source != null)
             {
                 splashVideo = string.Format(videoPathTemplate, "sources", game.Source.Id.ToString());
                 if (FileSystem.FileExists(splashVideo))
                 {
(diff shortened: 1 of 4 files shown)


source/Generic/SplashScreen/SplashScreenSettingsView.xaml
```diff
@@ -23,13 +23,15 @@
                 <StackPanel Margin="40,0,0,0" IsEnabled="{Binding ElementName=ExecuteDesktopMode, Path=IsChecked}">
                     <CheckBox IsChecked="{Binding Settings.ViewImageSplashscreenDesktopMode}" Content="{DynamicResource LOCSplashScreen_SettingCBviewImageSplashscreenDesktopMode}" Margin="0,10,0,0" />
                     <CheckBox IsChecked="{Binding Settings.ViewVideoDesktopMode}" Content="{DynamicResource LOCSplashScreen_SettingCBviewVideoDesktopMode}" Margin="0,10,0,0" />
+                    <CheckBox IsChecked="{Binding Settings.UseMicroTrailersInDesktopMode}" Content="{DynamicResource LOCSplashScreen_SettingCBuseMicroTrailerDesktopMode}" Margin="0,10,0,0" />
```
darklinkpower, line 26: Some things:

1. Bind their IsEnabled behavior to their parent CheckBox to enable the video functionality
2. Add a left margin of 20 to visually indicate that it's a child element of the parent option
3. A single `Use MicroTrailers` localization string will suffice for both. It will be clear for which mode it is due to their parent option of each checkbox

source/Generic/SplashScreen/SplashScreen.cs
```diff
@@ -331,48 +331,59 @@ private void SplashWindowClosed(object sender, EventArgs e)
 
         private string GetSplashVideoPath(Game game)
         {
-            var videoPathTemplate = Path.Combine(PlayniteApi.Paths.ConfigurationPath, "ExtraMetadata", "{0}", "{1}", "VideoIntro.mp4");
+            const string videoIntroName = "VideoIntro.mp4";
```
darklinkpower, line 334: Keep using `var` for consistency or make it a constant property outside of the method, whichever is fine

source/Generic/SplashScreen/SplashScreen.cs
```diff
@@ -331,48 +331,59 @@ private void SplashWindowClosed(object sender, EventArgs e)
...
             {
-                splashVideo = string.Format(videoPathTemplate, "playnite", "Fullscreen");
+                splashVideo = string.Format(videoPathTemplate, "platforms", platform.Id.ToString());
+                if (FileSystem.FileExists(splashVideo))
+                {
+                    return LogAcquiredVideo($"Platform '{platform.Name}'", splashVideo);
+                }
             }
 
+            var mode = isDesktop
```
darklinkpower, line 378: I think it's fine to keep it in a single line as it's not long enough to require multiple lines

darklinkpower (changes requested):

source/Generic/SplashScreen/SplashScreenSettingsView.xaml
cnapolit: Added name to parent & bound child enabled status to parent and applied other recommended changes.

source/Generic/SplashScreen/SplashScreen.cs
cnapolit: Updated.

source/Generic/SplashScreen/SplashScreen.cs
cnapolit: Moved to other constants

source/Generic/SplashScreen/SplashScreen.cs
```diff
@@ -381,6 +392,17 @@ private string GetSplashVideoPath(Game game)
             }
         }
 
+        private static string LogAcquiredVideo(string source, string videoPath)
+        {
+            logger.Info($"{source} video found in {videoPath}");
+            return videoPath;
+        }
+
+        private bool ShouldUseMicroTrailer(bool isDekstop)
```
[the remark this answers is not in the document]
cnapolit: Fixed

cnapolit:
- [ ] Squash to single commit

PR #447, opened 2023-10-22
[PlayNotes] Use font family and size from Playnite settings

EricKotato:
### The problem

Notes from PlayNotes plugin uses Calibri font instead of Playnite's font:

> Theme: Mythic, UI font: IBM Plex Sans, but font in screenshot is certainly wrong.

### The solution

The thing is, font settings are taken from MdXaml's Markdown style. Initially I've found the way to fix it from XAML (ref: https://github.com/whistyun/MdXaml/issues/14), but since it uses binding, the fix was need from C# side.

After adding the first line the font looked good, but the size was wrong:

> Theme: Mythic, UI font: IBM Plex Sans. Font is same, but it's even bigger than tab title.

This has been solved by adding the second line, where FontSize property is taken from settings:

> Theme: Mythic, UI font: IBM Plex Sans. Everything looks good now.

### Additional possible fixes

It's possible that font color could be also adjusted to themes. Here's an example on Mythic theme:

To do this, I've added a follwing line after proposed ones:
```c#
markdownStyle.Setters.Add(new Setter(FlowDocument.ForegroundProperty, ResourceProvider.GetResource("TextBrushDark")));
```

However, I'm not sure that using `TextBrushDark` color will suite any theme, so I haven't introduced it in this PR.

---
I have verified that:

### DIFF
--- a/source/Generic/PlayNotes/PlayNotesSettings.cs
+++ b/source/Generic/PlayNotes/PlayNotesSettings.cs
@@ -85,6 +85,29 @@ namespace PlayNotes
                 imageStyle.Setters.Add(new Setter(Image.MaxWidthProperty, maxWidthBinding));
             }
 
+            var baseStyleName = plugin.PlayniteApi.ApplicationInfo.Mode == ApplicationMode.Desktop ? "BaseTextBlockStyle" : "TextBlockBaseStyle";
+            if (ResourceProvider.GetResource(baseStyleName) is Style baseStyle &&
+                baseStyle.TargetType == typeof(TextBlock))
+            {
+                foreach(Setter s in baseStyle.Setters.Cast<Setter>())
+                {
+                    switch (s.Property.Name)
+                    {
+                        case "FontSize":
+                            markdownStyle.Setters.Add(new Setter(FlowDocument.FontSizeProperty, s.Value));
+                            break;
+
+                        case "FontFamily":
+                            markdownStyle.Setters.Add(new Setter(FlowDocument.FontFamilyProperty, s.Value));
+                            break;
+
+                        case "Foreground":
+                            markdownStyle.Setters.Add(new Setter(FlowDocument.ForegroundProperty, s.Value));
+                            break;
+                    }
+                }
+            }
+
             settings.MarkdownStyle = markdownStyle;
         }
 


source/Generic/PlayNotes/PlayNotesSettings.cs
```diff
@@ -84,6 +84,8 @@ public PlayNotesSettingsViewModel(PlayNotes plugin)
                 maxWidthBinding.RelativeSource = relativeSource;
                 imageStyle.Setters.Add(new Setter(Image.MaxWidthProperty, maxWidthBinding));
             }
+            markdownStyle.Setters.Add(new Setter(FlowDocument.FontFamilyProperty, ResourceProvider.GetResource("FontFamily") as FontFamily));
+            markdownStyle.Setters.Add(new Setter(FlowDocument.FontSizeProperty, ResourceProvider.GetResource("FontSize")));
```
darklinkpower, line 88: Have you checked if both resources are available in Fullscreen Mode too? I haven't touched Fullscreen mode development so I'm not sure if it uses the same resources or not.

darklinkpower:
Regarding the text, it would be better to use `TextBrush` because that's the standard text color but if you add it, please add a null check in case the theme doesn't have that resource and I think should also be casted for safety. An example of how to do this is in the Notes control, using the `is` operator: https://github.com/darklinkpower/PlayniteExtensionsCollection/blob/e16e3902d0c0caa67f9e9257267229e827cc9fca/source/Generic/PlayNotes/PlayniteControls/NotesViewerControl.xaml.cs#L176

source/Generic/PlayNotes/PlayNotesSettings.cs
darklinkpower: It seems it doesn't: https://github.com/JosefNemec/Playnite/blob/4c03306ff4bf62ed2a244b20543f2839eb5483cf/source/Playnite.FullscreenApp/Themes/Fullscreen/Default/Constants.xaml#L37-L40

Apparently it uses `FontTitilliumWebLight` but perhaps it would make sense to get the `TextBlockBaseStyle` and grab its `FontFamily`, just in case it was changed in the theme.

darklinkpower: You can use `ApplicationInfo.Mode` in `plugin.PlayniteApi` to obtain the Playnite mode.

EricKotato: Is there a fullscreen theme which supports Play Notes plugin? I've made some changes, and I'd like to test them.

EricKotato: I've tried to test the plugin by injecting it in default fullscreen Playnite theme. It looks that it has multiple problems.



1. All icons are missing.
2. Base control is unstyled, which forces background to be white. I've tried to inject TabControl.xaml from default theme, but it didn't work.

But if you'll select the font, it looks like it's correct:


The font is bigger than description though, but the problem is that in default theme the description size is hard-coded:

https://github.com/JosefNemec/Playnite/blob/4c03306ff4bf62ed2a244b20543f2839eb5483cf/source/Playnite.FullscreenApp/Themes/Fullscreen/Default/Views/GameDetails.xaml#L118-L122

Even though the new code works, more work is needed to be actually usable in fullscreen mode.

source/Generic/PlayNotes/PlayNotesSettings.cs
```diff
@@ -85,6 +85,28 @@ public PlayNotesSettingsViewModel(PlayNotes plugin)
                 imageStyle.Setters.Add(new Setter(Image.MaxWidthProperty, maxWidthBinding));
             }
 
+            if (ResourceProvider.GetResource(baseStyleName) is Style baseStyle &&
```
darklinkpower, line 88: Did you build and test the changes? Correct me if I'm wrong but `baseStyleName` is not defined anywhere in this scope

source/Generic/PlayNotes/PlayNotesSettings.cs
darklinkpower: The reason is that FS mode doesn't have styling for TabControl by default but this change was required to discard the possibility of crashing on startup

source/Generic/PlayNotes/PlayNotesSettings.cs
EricKotato: Forgot the line, sorry, fixed.

darklinkpower: I see, no worries. Thanks for contributing.

Issue #469, opened 2023-11-18
[Game Relations] Games without a category or tag or genre won't show up in similar games


HerrKnarz:
### Extension name

Game Relations

### Bug Description

Since version 1.3 of Game Relations I noticed, that I don't see any similar games anymore, although I have them activated. It looked into the settings and tried a few things. The reason was, that none of my games have a category, since I don't use categories at all. After adding some games to a test category, they had similar games, but only those, that also have the same category. After removing the category from a game, no similar games were showing up anymore.

I also tried putting that test category to the excluded list to see, if games without that category now have similar games, but it seems, the addon only takes games into consideration, that have at least one category.

After some more tests, it looks like the same bug occurs with tags and genres.

### To Reproduce

Look for a game, that has similar games, then remove all categories or tags or genres from that game. Switch to another game and back. Now there won't be any similar games anymore.

### Logs

[playnite.log](https://github.com/darklinkpower/PlayniteExtensionsCollection/files/13400319/playnite.log)
[extensions.log](https://github.com/darklinkpower/PlayniteExtensionsCollection/files/13400320/extensions.log)

darklinkpower:
This is by design. From my tests the best way to match is by using Tags, Genres and categories at once or otherwise many games that are not very similar could show up. Still, the algorithm could be improved to be more reliable when any of these fields are missing.

PR #505, opened 2024-02-18
Fixed 2 games both with empty collections resulting in 0 similarity (#469)

Jeshibu:
I have verified that:

### DIFF
--- a/source/Generic/GameRelations/PlayniteControls/GameRelationsBase.xaml.cs
+++ b/source/Generic/GameRelations/PlayniteControls/GameRelationsBase.xaml.cs
@@ -9,19 +9,10 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
 using System.Runtime.CompilerServices;
-using System.Text;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
-using System.Windows.Data;
-using System.Windows.Documents;
-using System.Windows.Input;
-using System.Windows.Media;
-using System.Windows.Media.Imaging;
-using System.Windows.Navigation;
-using System.Windows.Shapes;
 using System.Windows.Threading;
-using TemporaryCache;
 
 namespace GameRelations.PlayniteControls
 {
@@ -213,7 +204,7 @@ namespace GameRelations.PlayniteControls
             }
 
             // Rounding is done to prevent errors when doing arithmetic operations
-            var matchPercent = Math.Round(commonCount / (decimal)Math.Max(listToMatch.Count(), hashSet.Count), 3); 
+            var matchPercent = Math.Round(commonCount / (decimal)Math.Max(listToMatch.Count(), hashSet.Count), 3);
             return matchPercent;
         }
 
@@ -226,8 +217,14 @@ namespace GameRelations.PlayniteControls
         /// <returns>A match value between 0 and 1 representing the similarity between the elements.</returns>
         protected static double CalculateJaccardSimilarity<T>(IEnumerable<T> listToMatch, HashSet<T> hashSet)
         {
-            if (listToMatch is null || !listToMatch.Any() || hashSet is null || hashSet.Count == 0)
+            if (listToMatch is null || !listToMatch.Any())
             {
+                //If both games have 0 items in this collection, give some pity similarity
+                if (hashSet is null || hashSet.Count == 0)
+                {
+                    return 0.9D;
+                }
+
                 return 0;
             }
 


darklinkpower:
Changes like this should be done in the control game matching algorithm and not to that method, since it calculates the [Jaccard Similarity](https://en.wikipedia.org/wiki/Jaccard_index) and this change would make it inaccurate. I'm also not sure if the best approach is to give a fixed number in such cases but I'm open to suggestions to improve the algorithm and if you've checked that it works fine.

Jeshibu:
I've changed it so that it only uses fields that have values for at least one of the games.
Fields with 0 (non-ignored) items in both lists now do not contribute to the similarity score, but also don't contribute to the minimum similarity score threshold.

I've also tweaked the minimum similarity threshold, but I want to put that in an advanced settings tab, along with the field weights, if you think this is the right way to go.

darklinkpower (changes requested):

source/Generic/GameRelations/PlayniteControls/SimilarGamesControl.cs
```diff
@@ -1,70 +1,55 @@
...
-using TemporaryCache;
 
 namespace GameRelations.PlayniteControls
 {
     public partial class SimilarGamesControl : GameRelationsBase
     {
-        private readonly Dictionary<string, double> _propertiesWeights;
-        private const double _minMatchValueFactor = 0.73;
+        private readonly Dictionary<GameField, double> _propertiesWeights;
+        private const double _minMatchValueFactor = 0.69;
```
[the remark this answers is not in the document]
Jeshibu: I don't think 0.73 is universal, especially if you tag a lot of incidental details like locations and characters.

Jeshibu:
Added settings. If this is about right I'll add localization too.

source/Generic/GameRelations/GameRelationsSettings.cs
```diff
@@ -185,6 +189,27 @@ private void RemoveSelectedItemsFromExclude(ObservableCollection<DatabaseObject>
...
 
+        private string GetResourceString(string key) => plugin.PlayniteApi.Resources.GetString(key);
+
+        public void SetAdvancedSectionDefaults()
+        {
+            Settings.SimilarGamesControlSettings.JacardSimilarityPerField = 0.73D;
+            Settings.SimilarGamesControlSettings.FieldSettings.Clear();
+            settings.SimilarGamesControlSettings.FieldSettings.Add(new SimilarGamesFieldSettings(GameField.TagIds, GetResourceString("LOCTagsLabel"), true, 1));
+            settings.SimilarGamesControlSettings.FieldSettings.Add(new SimilarGamesFieldSettings(GameField.GenreIds, GetResourceString("LOCGenresLabel"), true, 1.2));
+            settings.SimilarGamesControlSettings.FieldSettings.Add(new SimilarGamesFieldSettings(GameField.CategoryIds, GetResourceString("LOCCategoriesLabel"), true, 1.3));
```
darklinkpower, line 200: I don't think the names should be set during their creation. This creates the issue that if language or source string changes, they'll stay the previous incorrect name so it would be better to resolve at runtime.

darklinkpower, line 200: I don't think the names should be set during their creation. This creates the issue that if language or source string changes, they'll stay the previous incorrect name so it would be better to resolve at runtime.

darklinkpower: Additionally, if the settings for each field are being made separate here, in my opinion it would make sense to also move the item exclusion functionality there too:

image

source/Generic/GameRelations/GameRelationsSettings.cs
```diff
@@ -185,6 +189,27 @@ private void RemoveSelectedItemsFromExclude(ObservableCollection<DatabaseObject>
...
+        }
+
+        public void UpgradeSettings()
+        {
+            int currentVersion = 2;
+
+            if (settings.SettingsVersion < 2)
+                SetAdvancedSectionDefaults();
```
darklinkpower, line 208: Please add curly braces here.

source/Generic/GameRelations/GameRelationsSettingsView.xaml
```diff
@@ -221,7 +221,40 @@
                                 </DockPanel>
                             </Grid>
                         </Expander>
-                        
+
+                        <Expander Margin="0,10,0,0" Header="Advanced">
+                            <StackPanel>
+                                <StackPanel Orientation="Horizontal" Margin="0,10,0,0">
+                                    <TextBlock VerticalAlignment="Center">Jacard similarity requirement per field:</TextBlock>
```
darklinkpower, line 228: Missing translation string

darklinkpower: I think the string should remain only as "Minimum similarity to match games" with a more detailed explanation in a tooltip, including saying that it uses Jacard.

source/Generic/GameRelations/GameRelationsSettingsView.xaml
```diff
@@ -221,7 +221,40 @@
...
+                                    <TextBox VerticalAlignment="Center" IsReadOnly="True" MinWidth="35" Margin="10,0"
+                                             Text="{Binding Settings.SimilarGamesControlSettings.JacardSimilarityPerField}" />
+                                    <Slider Minimum="0.4" Maximum="1" IsSnapToTickEnabled="True" TickFrequency="0.01"
+                                            Value="{Binding Settings.SimilarGamesControlSettings.JacardSimilarityPerField}"
+                                            Width="200" VerticalAlignment="Center"/>
+                                </StackPanel>
+
+                                <TextBlock Text="Fields enabled/weights:" Margin="0,10,0,0"/>
```
darklinkpower, line 236: I don't like much the text here. Perhaps it should say "Individual fields settings"

source/Generic/GameRelations/PlayniteControls/GameRelationsBase.xaml.cs
```diff
@@ -284,7 +275,7 @@ protected static T GetAnyCommonItem<T>(List<T> listToMatch, HashSet<T> hashSet)
         /// Returns false if the list is null or empty.</returns>
         protected static bool HashSetContainsAnyItem<T>(IEnumerable<T> listToMatch, HashSet<T> hashSet)
         {
-            if (listToMatch is null || !listToMatch.Any())
+            if (listToMatch is null)
```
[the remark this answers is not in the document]
darklinkpower: This was erroneously marked as resolved without restoring an original condition.

source/Generic/GameRelations/PlayniteControls/SimilarGamesControl.cs
```diff
@@ -1,70 +1,55 @@
...
+            var gtm = GetGameToMatchInfo(game);
 
-            var seriesHashSet = game.SeriesIds?.ToHashSet() ?? new HashSet<Guid>();
-
-            var minScoreThreshold = _propertiesWeights.Count * _minMatchValueFactor;
             var similarityScores = new Dictionary<Game, double>();
             foreach (var otherGame in PlayniteApi.Database.Games)
             {
-                if (otherGame.Id == game.Id)
+                if (otherGame.Id == gtm.Game.Id)
```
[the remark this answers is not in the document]
darklinkpower: Same. Please don't use acronyms.

source/Generic/GameRelations/PlayniteControls/SimilarGamesControl.cs
```diff
@@ -73,5 +64,70 @@ public override IEnumerable<Game> GetMatchingGames(Game game)
 
             return similarGames;
         }
+
+        private GameToMatchInfo GetGameToMatchInfo(Game game)
+        {
+            var gtm = new GameToMatchInfo(game);
```
darklinkpower, line 70: Same.

darklinkpower (changes requested):

PR #592, opened 2024-11-25
update endpoints for Jast

christopherhero:
- generate shop token
- get user games

I have verified that:
- [ ] These changes work, by building the extension and testing.
- [ ] That the changes comply with the [rules](https://github.com/darklinkpower/PlayniteExtensionsCollection#contributing) indicated in the repository.
- [ ] Pull request is targeting `master` branch.

### DIFF
--- a/source/Library/JastUsaLibrary/JastUrls.cs
+++ b/source/Library/JastUsaLibrary/JastUrls.cs
@@ -14,14 +14,14 @@ namespace JastUsaLibrary
             public static class Authentication
             {
                 private const string PathBase = "shop/";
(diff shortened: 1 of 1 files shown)


christopherhero:
@darklinkpower in some time there will be a change of endpoint names in the jastUSA api

darklinkpower:
Thanks for the PR. Could you clarify where this information about the API changes is coming from and how you know the endpoints will change? Is there a specific source or way to stay updated on future changes?

I'm currently in the process of (slowly) refactoring the extension, so if you have an approximate time frame for when the change will happen, it would be really helpful. That way, I can aim to complete the refactor by then.

Issue #706, opened 2025-12-06
[Extra Metadata Loader] Doesn't compile on main branch


supersmo:
### Pre-Request Checklist

- [x] I have checked and confirmed **this is not** a duplicate issue
- [x] I have verified that I'm using the latest version of the extension by checking for updates in Playnite
- [x] I have modified the placeholder issue title according to this issue

### Extension name

Extra Metadata Loader

### Bug Description

I am unable to build Extra Metadata Loader.

For instance:
[ExtraMetadataLoader.cs line 904](https://github.com/darklinkpower/PlayniteExtensionsCollection/blob/a3a9f36c73ca1e8e4074ef13e4d06758cdaced6f/source/Generic/ExtraMetadataLoader/ExtraMetadataLoader.cs#L904)
```
var downloadOptions = new VideoDownloadOptions(VideoType.Trailer,);
```
I can't find the definition for `VideoType.Trailer` and the constructor for VideoDownloadOptions doesn't take an enum. 

I also see a similar issue for calls to `GetGameLogo`. I can't find the definition for that method.

Could the main branch happen to have work in progress refactoring changes?

### Steps to Reproduce

Clone the main branch and build in Visual Studio 2019 v16.x

### Logs

I currently have 18 errors in the error list but these lines are related to the example:

The name 'VideoType' does not exist in the current context

'VideoDownloadOptions' does not contain a definition for 'VideoType' and no accessible extension method 'VideoType' accepting a first argument of type 'VideoDownloadOptions' could be found (are you missing a using directive or an assembly reference?)

darklinkpower:
The plugin source in the main branch is a work in progress rewrite. The current and build-able source is in the `emlCrashFix2025` branch

Issue #705, opened 2025-11-27
[ExtraMetadataLoader] Stopped fetching steam trailers for newly added steam games


majima-everywhere:
### Pre-Request Checklist

- [x] I have checked and confirmed **this is not** a duplicate issue
- [x] I have verified that I'm using the latest version of the extension by checking for updates in Playnite
- [x] I have modified the placeholder issue title according to this issue

### Extension name

ExtraMetadataLoader

### Bug Description

Since recently (26th Nov), ExtraMetadataLoader no longer retrieves Steam video trailers for EVERY newly added games in the Playnite library. 

What still works:
- Video trailers continue to appear and play correctly for older games already in the library before the issue began
- Steam logos (auto or manual retrieval) still work normally for both old and new games

Even if a game already exists in my playnite library with a working steam video trailer, creating a new second entry for the same game causes the new entry to fail to display any trailer, while the old entry continues to work normally.

### Steps to Reproduce

1. Add a steam game / or (manually) a new game  and retrieve its steam metadata
2. Open the game’s detail page
3. The entry doesn't have a "video" section (The logo loads and appears successfully though)
4. Check extensions.log `|ERROR|ExtraMetadataLoader#VideoPlayerControl:Error deserializing steam appdetails json file in X:\Playnite\ExtensionsData\705fdbca-e1fc-4004-b839-1d040b8b4429\a5e572b7-1ee0-4298-94b9-c9d97945c219_SteamAppDetails.json`

_The game is Hi-Fi rush picked randomly for demonstration purposes_

(Optional) Create a duplicate entry of a game's playniote page that already has a working trailer => The old entry works while the new entry fails.

### Logs

Error shown when opening a newly created game entry:
```
|ERROR|ExtraMetadataLoader#VideoPlayerControl:Error deserializing steam appdetails json file in X:\Playnite\ExtensionsData\705fdbca-e1fc-4004-b839-1d040b8b4429\a5e572b7-1ee0-4298-94b9-c9d97945c219_SteamAppDetails.json
|DEBUG|Extras#ThemeExtras:Pushed "Grid View - Hi-Fi RUSH - DEMO BUG"
```

Opening the old (working) entry produces no errors:
`|DEBUG|Extras#ThemeExtras:Pushed "Grid View - 🎮 Hi-Fi RUSH - Steam"`

### Screenshots

newly created game entry:
[...]

PR #707, opened 2025-12-07
Guess Steam mp4 trailer url if not found

supersmo:
I have verified that:
- [ ] Pull request is targeting `master` branch.

Pull request is not targeting the `master` branch since it is not the current working branch.
See: https://github.com/darklinkpower/PlayniteExtensionsCollection/issues/706#issuecomment-3621407679

This changes the behavior for Steam trailer on demand streaming. Steam recently stopped returning the mp4 URL of trailers in the `appdetails` response.
* If the mp4 url is available in the cached `appdetails` file then that is used.
* Otherwise an attempt is made to guess the URL of the mp4.
* A simple test is made to see if the URL is available and if it is then that is used to display the trailer.

This is a temporary solution. A more robust solution would be to get a video player to play the m3u8 or mpd urls of the trailers.

See: https://github.com/darklinkpower/PlayniteExtensionsCollection/issues/705

### DIFF
--- a/source/Generic/ExtraMetadataLoader/Controls/VideoPlayerControl.xaml.cs
+++ b/source/Generic/ExtraMetadataLoader/Controls/VideoPlayerControl.xaml.cs
@@ -20,6 +20,7 @@ using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
 using System.Windows.Threading;
+using System.Net.Http;
 
 namespace ExtraMetadataLoader
 {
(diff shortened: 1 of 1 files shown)


justin-delano:
can you provide a .pext for testing? I've run into the same issue!

joyrider3774:
i compiled and tried this, but when i right click on a game a choose to download video trailers it does not find any, however it does find the microtrailers it seems

justin-delano:
@supersmo thank you so much, this worked for me!

PR #740, opened 2026-04-19
Streaming fallback

tetj:
I have verified that:


tetj:
My changes are here : https://github.com/darklinkpower/PlayniteExtensionsCollection/pull/740/commits/a9171d2bbb9ee3028547a9f173f6a7732ca87e6e

**TLDR :** The goal of this feature is to have a "streaming fallback" so when a video can't be streamed from Steam, it is automatically downloaded on the fly. 

**In other words :** Per-game automatic fallback in the player: In VideoPlayerControl.UpdateGameVideoSources, when StreamSteamVideosOnDemand is enabled but the game has no streamable Steam movie (e.g., non-Steam game, or Steam appdetails returns no movies), automatically trigger a one-off Steam video download for that game and play it once it's on disk. 

It takes less than 10 seconds on my PC for the video to download in the background. It does not really slow down the user since the download is invisible to the user and the app is usable during the download. It helps saving disk space. I had over 100GB of video trailers on my disk because I was using "Download videos of newly added games on library update". It also makes library update much faster.

PR #766, opened 2026-07-26
JAST USA Library: Handle updated product response

### LINKED ISSUE #765: [JAST USA Library] Error During Library Update
### Pre-Request Checklist

- [x] I have checked and confirmed **this is not** a duplicate issue
- [x] I have verified that I'm using the latest version of the extension by checking for updates in Playnite
- [x] I have modified the placeholder issue title according to this issue
- [x] I have attached the requested logs, and I understand that issues submitted without logs may not be investigated and may even be closed

### Extension name

JAST USA Library

### Bug Description

I consistently get the following error when updating my library since a few days ago:

Image

I suspect this may have something to do with the recent updates to their website. They recently migrated everything and all the old links are now redirects.

### Steps to Reproduce

Attempt to update the library. That's about it.

### Logs

[playnite.log](https://github.com/user-attachments/files/30360951/playnite.log)
[extensions.log](https://github.com/user-attachments/files/30360952/extensions.log)

### Screenshots

_No response_

Dawnflare:
## Summary
Fixes the JAST failed sync issue described in #765 

## Details
- update the JAST account response DTO to ignore the now-reference-only plural `variants` field and deserialize `game` from the containing product
- retain complete products while paging through the account library response
- map library entries from the embedded singular `variant` plus product-level `game` data
- preserve compatibility with the previous response layout by falling back to `variant.game`

## Root cause

JAST changed the authenticated library response around July 23, 2026. The plural `products[].variants` field now contains JSON-LD URI references such as `/api/v2/shop/product-variants/nnya033` instead of embedded variant objects. The existing DTO declared this field as `Variant[]`, so Newtonsoft.Json aborted with a `JsonSerializationException` before the library could be imported.

The same response update also moved `game` from the singular embedded `variant` to its containing product. Ignoring only the plural field would therefore deserialize successfully but later fail when mapping the game and translation data.

## User impact

JAST USA library synchronization currently fails for affected users on every attempt. This change accepts the updated response shape and restores game import while remaining compatible with the earlier nested-game shape.

## Validation

- rebuilt `JastUsaLibrary.csproj` successfully with MSBuild in Debug configuration
- deserialized a captured current JAST response containing 20 products (120 total across 6 pages); all products retained the required variant and game data
- verified a legacy response fixture still resolves `game` through the fallback path
- installed the Debug build into a Playnite 10 portable installation and completed a real authenticated JAST library sync successfully
- ran `git diff --check`

Fixes #765

### DIFF
--- a/source/Library/JastUsaLibrary/Services/JastUsaIntegration/Application/Services/JastUsaAccountClient.cs
+++ b/source/Library/JastUsaLibrary/Services/JastUsaIntegration/Application/Services/JastUsaAccountClient.cs
@@ -96,16 +96,20 @@ namespace JastUsaLibrary.JastUsaIntegration.Application.Services
 
             var products = await _apiClient.GetProductsAsync(token, cancellationToken);
             var gamesData = products.Select(product =>
-                new JastGameData(
-                    product.ProductName,
-                    product.ProductCode,
-                    product.GameId,
-                    product.Game.ApiRoute,
-                    product.Game.Translations.EnUs?.Id,
-                    product.Game.Translations.Ja?.Id,
-                    product.Game.Translations.ZhHans?.Id,
-                    product.Game.Translations.ZhHant?.Id
-                )).ToList();
+            {
+                var variant = product.Variant;
+                var game = product.Game ?? variant.Game;
+                return new JastGameData(
+                    variant.ProductName,
+                    variant.ProductCode,
+                    variant.GameId,
+                    game.ApiRoute,
+                    game.Translations.EnUs?.Id,
+                    game.Translations.Ja?.Id,
+                    game.Translations.ZhHans?.Id,
+                    game.Translations.ZhHant?.Id
+                );
+            }).ToList();
 
             return gamesData;
         }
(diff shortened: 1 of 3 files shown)


darklinkpower:
Hi, thanks for the PR. I've implemented a fix in https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/00d2710ecd68b95f70a96e7a6834679595e47c69 but I noticed that there are some differences in your PR. 

In here I noticed that you fallback to `variant.Game`

image

But in the response I got to implement the fix, `variant` object does not have a `Game` field. The only object that does is `Product`. Did you see that in your response? And for you, were there cases where certain Products didn't have a `Game` field?

Dawnflare:
Thank you for checking.  I apologize for the delayed response.  I was out of town on vacation.

In the current response I inspected, I saw the same as you did: `Game` was present on `Product`, not on `variant`.

In the response captured in the logs attached to #765, all 20 products on the logged first page had `Product.Game`, and none of their embedded `variant` objects had a `Game` field. I did not observe any products missing `Product.Game`.

The `product.Game ?? variant.Game` fallback was only a defensive compatibility measure based on the previous DTO, which modeled `Game` under `Variant`. It was not based on seeing `variant.Game` in the new response. Given the current schema and your testing, I agree that the fallback is unnecessary.

Since your fix in [`00d2710`](https://github.com/darklinkpower/PlayniteExtensionsCollection/commit/00d2710ecd68b95f70a96e7a6834679595e47c69) has now been tested and released and also addresses metadata downloads, this PR has been superseded. I’ll close it. 

Thank you for implementing and releasing the fix!

darklinkpower:
Thanks for looking into it and the reply, I really appreciate it! Hopefully the fix is working well for all users and there aren't any surprises.

PR #371, opened 2023-02-10
Added workshop link

D1yt:
Checks if workshop link metadata exists and only creates button if it does. Link from metadata is passed to this extension.

I have verified that:
- [X] These changes work, by building the extension and testing.
- [X] That the changes comply with the [rules](https://github.com/darklinkpower/PlayniteExtensionsCollection#contributing) indicated in the repository.
- [X] Pull request is targeting `master` branch.

PR #767, opened 2026-08-06
Add workshop link for Steam Shortcuts

### LINKED ISSUE #391: [Steam Viewer] Add option to open game workshop
### Extension name

Steam Viewer

### Feature Description

Add an option to the dropdown that allows users to open the selected game workshop

### Screenshots

_No response_

CarrieForle:
I have verified that:

This PR adds workshop link in the menu of Steam Shortcuts. Workshop item uses a new localization string `LOCSteam_Viewer_MenuItemGameWorkshopDescription` that defaults to "Workshop".

Merging this allows to close #391 and obsolete PR #371. Unlike #371 however, it always adds workshop item in the menu regardless whether the game supports it or not.


darklinkpower:
> however, it always adds workshop item in the menu regardless whether the game supports it or not.

I'm sorry but this is not a good approach. If implemented, it should properly detect if the Workshop is available for the specific game. Additionally, the localization string must only be added to the `en_US.xaml` file but not to the other ones.

CarrieForle:
> the localization string must only be added to the en_US.xaml file but not to the other ones.

I've updated it to remove the string in those xaml files.

> If implemented, it should properly detect if the Workshop is available for the specific game

I can try to implement it, but if that's the case then achievements and points shop also deserve the same treatment as they are also optional.

Examples:
- No achievement: https://store.steampowered.com/app/3241660/REPO/
- No points shop: https://store.steampowered.com/app/1205170/Super_Magbot/

CarrieForle:
I've made it so the workshop is hidden if the game does not support it. I also made this work for achievements and points shop so they would be hidden as well if the game doesn't have it. As  this is no longer just about workshop please tell me if this needs to be in another PR.

### Implementation

In order to get support info I use https://api.steampowered.com/ILoyaltyRewardsService/QueryRewardItems/v1/?appids[0]={0}&count=1 to get points shop by checking `total_count > 0` and `SteamWeb.GetSteamAppDetails()` to get achievements and workshop.

The requests take a bit of time so I send the requests early when the user selects a game, so there is no delay between menu appearance and button clicks. In cases where the user right clicks on the game before they select it, all the links are shown whether they are supported or not because I couldn't find a timing to do it early, and UI would be very unresponsive if the requests were sent after button click.

